PR #161, opened 2018-10-02
160 show grid

### LINKED ISSUE #160: Show Grid Component
# Description of Issue

This issue replaces issues #140 and #142 - we can use one component for both "my shows" and "my past shows". 

# Possible Solution and/or Approach

I'll make a component called ShowGrid which will display show informations

ogdenstudios:
# Issue

This issue resolves #160 

# Description

This PR fleshes out the components for ShowTable, its JSX, and empty SCSS classes. The ShowTable component includes a TBody and THead component, and we style with classes for .table, .table__header, .table__header__cell, .table__body, .table__row, and .table__cell 

Right now the css for these selectors are empty, pending design changes, but are applied to all elements. 

I popped in a couple of headers and two ShowTable components on the Dashboard page for an "Upcoming Shows" and "Past Shows" render.


dzimmerman902:
Hey Tyler! So as I normally do, I go back a forth on how I want to handle things. I'm trying to simplify the process at much as possible 

- I don't think it's necessary to give all the table components classes. The only time we would use a tr, td, etc. is within a table. So using standard css notation where we can jump to child elements will be sufficient.
- Instead of creating separate folders for each table and creating separate components for each table, I think it's best to go ahead and use one file to handle all of the table components. 
- Since the goal is to utilize Redux, so we can make them class components so we can grab teh data that way and not worry about passing props. I did it for these anyways for now until we figure out how we are getting the data. 

We'll talk more and make sure you're cool with the changes. Thanks!

ogdenstudios:
Yea that all sounds good to me. I'll check in tonight sounds straightforward. It looks like you made the changes here. Do you still need me to do anything on this PR or are your changes pretty much what you want in it for now?

dzimmerman902:
We'll go over tonight. Want to explain the method to my madness before I merge.

PR #199, opened 2018-10-23
Add Users Seed Script

ogdenstudios:
# Description

I generated some seed data - 50 randomly generated users in scripts/seedDB/lib/users.json and scripts/seedDB/users.js seeds them to the database without blowing away the comrad.development account, pretty reliably. 

Album seeding still doesn't work, and you'll get errors when running npm run seed, but the users will still get created. 

Originally wanted to run this with something like 1000+ users, but ran into issues with the mongoose buffer. We can likely fix this using [Mongoose bufferCommands parameter](https://mongoosejs.com/docs/guide.html#bufferCommands).

For now, 50 seed users should give us the ability to work on the users API.


ogdenstudios:
Closing this to make way for @msiadak's PR first, then redoing on top of his refactoring.

PR #211, opened 2018-10-28
Improve music library keyword search

### LINKED ISSUE #50: Keyword search of library
Search for library of tracks via some API

msiadak:
# Issue

This issue resolves #50 

# Description

This PR improves the searchLibrary controller function, adding the following capabilities:

* Splits the query input string into individual words and searches for a match in any of them
* Searches child documents (e.g. album artist, track artist, track album) for text matches
* Calculates relevance of search results -- if the query string matches the 'name' field of the document exactly, relevance is 100.  Otherwise, relevance is the number of word matches found in all searched fields.

There are better, more elegant ways to do search relevance, but I tried to keep it simple for the first attempt at this.  Later, we may want to look into adding something like ElasticSearch to handle this.  This may be especially necessary as the size of the music library increases.  With this approach we're doing a lot of MongoDB queries and iterating over the results with a bunch of RegExp matching, so I think the time complexity is pretty high.

Sorry for the commit mess -- I messed something up while trying to rebase onto development.  Should be fixed if we squash+merge.

### DIFF
--- a/controllers/searchController.js
+++ b/controllers/searchController.js
@@ -32,6 +32,25 @@ function filterDuplicates(results) {
   ).filteredResults;
 }
 
+/**
+ * Queries the Album, Artist, and Track collections for a given query string
+ * and populates the relationships in Track.
+ */
+async function searchLibrary(query) {
+  const queryRegexp = new RegExp(query, 'i');
+  const resultSets = await Promise.all([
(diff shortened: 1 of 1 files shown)


dzimmerman902:
Looks really good!

PR #213, opened 2018-10-29
111 get users endpoint improvements

### LINKED ISSUE #111: Get Users endpoint improvements
The Get Users endpoint needs to:
- sort by Name by default
- have options to order by DJ Name or Active/Inactive status
- have a filter that could return only Active users, only Inactive users, or All users
- allow a search string, which will return results filtered by testing the search string against Name and DJ Name
- page items with 100 results per page

(see the wireframes at https://drive.google.com/open?id=13xQLl3P91aNjqmIs1vdUzmtxQkesDgDU -- the User List and Add/Edit User page will be how this is used on the front-end)

ogdenstudios:
# Issue

This issue resolves #111 

# Description

I believe I've resolved #111 with this PR. Here's how: 

1. I've added the ability for users API routes to accept a sort_by query. If it is not provided, all results will default to sorting by "on_air_name" - which is what I took the original issue to mean by "sort by name". That's an easy tweak if we want to sort by first or last name. If you hit `/api/user`, you'll get results sorted by "on_air_name". But you could hit` /api/user?sort_by=email` and get results sorted by email. You could use the sort_by query to order by Active/Inactive status, as well, by setting `sort_by=status`

2. Added filters for active/inactive users, which are, respectively: 

- `/api/user/filter/active`
- `/api/user/filter/inactive`

3. `api/user/search/` allows you to pass a string, and runs a search against on_air_name, first_name, and last_name in mongoose and returns a match based on that string. Right now it doesn't use regex, so it needs to be exact. Adding regex might make our search more powerful, but I'm unsure what we want to do about that. 

4. All of these endpoints allow us to pass in queries for `limit` and `page`. The limit defaults to 10, instead of 100, but we can change that easily. The page will default to 0. 

You can check out test API calls in Postman account to see how it all works.


dzimmerman902:
@ogdenstudios Great work on the filter feature. Since there is a lot of duplicate code, I would try to find a way to create one filter controller instead of having two separate ones.

ogdenstudios:
Good catch. I'll revise! Thanks!

dzimmerman902:
Nice, I'll review it and let you know what I think!

ogdenstudios:
I was going to fix the case sensitivity, but then realized I'm not sure if that's a bug or desired behavior. Same for the name searches: do we want to regex these? And if so, do we want to open a separate issue for that kind of thing? Seems like maybe a bigger concern than just for these specific API endpoints.

dzimmerman902:
Getting closer! We're definitely going to need an API route that searches the user just by ID, go ahead and leave the ID one in and update the findAll and findStatus

dzimmerman902:
Also, I think we're going to work with a separate controller when it comes to searching for multiple documents in a collection, we can talk more on Monday!

ogdenstudios:
Done! Hope this one does it. Thanks for all the code review!

PR #223, opened 2018-11-06
Add a UI page design to the readme

wdoug:
# Description

I've been adding these to all our projects so new folks can get a sense of what the projects are like. It also gives a reference to an image so I can include it in the members' project [here](https://members.codefordenver.org/projects)


dzimmerman902:
I like it!

PR #224, opened 2018-11-06
Moved server files to seperate folder

dzimmerman902:
## Guidelines

YOUR branch on the RIGHT

## Issue

No Issue

## Description

Re-organized the folder structure for the project. I put all the server files into one folder to keep the root of the project cleaner because I anticipate more files that will be needed in the root. Will make sure everyone is cool with the move.


dzimmerman902:
Also, I tried to move the server .env file into the server file but it didn't work. Keep that file in the root for it to work

ogdenstudios:
I like it and it runs just fine for me.

PR #225, opened 2018-11-07
Remove some junk text from readme

ogdenstudios:
I was testing out a markdown plugin in my editor, added a line to the readme, and it snuck in. Removing that minor mistake on my end.


dzimmerman902:
Does this need to be merged?

Issue #218, opened 2018-11-06
Alerts HTML/CSS


seankwilliams:
Design HTML/CSS for the alerts in:

https://docs.google.com/document/d/1A7SExuMphBB8WER-wt8fVyPN3o4fdvJkEvvs2fViVrQ/edit

seankwilliams:
Awesome, thanks @dzimmerman902 and good work @crodier1 !

PR #226, opened 2018-11-07
Add 401 component

### LINKED ISSUE #132: 401 Unauthorized Component
# Guidelines

Have a short descriptive title <br>
Make sure to click the following:

- [ ] Assignees
- [x] Labels
- [ ] Milestone

# Full Details

Would like a simple component to display is a user tries to access a page they are not authorized to get too.

ogdenstudios:
## Issue

This issue resolves #132 

## Description

Added an `Unauthorized` component for when user tries to access something they shouldn't. It's missing the alert from Lauren's design, but that's being completed in #218 - once that alert is ready, we should be able to drop it on in there.

### DIFF
--- a/client/src/components/Unauthorized/Unauthorized.js
+++ b/client/src/components/Unauthorized/Unauthorized.js
@@ -4,7 +4,7 @@ class Unauthorized extends Component {
   render() {
     return (
       <div>
-        <h1>401: Unauthorized Access</h1>
+        <h1 className="unauthorized__h1">401: Unauthorized Access</h1>
         <div>
         Alert button with unauthorized text goes here
         </div>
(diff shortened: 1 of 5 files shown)


dzimmerman902:
Great work @ogdenstudios! Let's try to think about how we can make this component more efficient.

Most likely we'll have different kinds of alerts with different colors/messages/etc. Try to create a more modular component that we can alter easier with props. 

For instance, if we pass in the props of danger, it will be red color. If we pass in the props of warning, it will be a yellow color.

Feel free to reach out if you have additional questions!

dzimmerman902:
@ogdenstudios My bad! I read through in more detail and I see the placeholder for the Alert component. You did the right thing!

I appreciate the great notes, my apologies for skipping over them!

dzimmerman902:
I'm going to check with the group on thoughts about doing a class or functional component for these types

ogdenstudios:
Not a problem, you are managing a lot here. Thanks for all the work you do! Catch you next Monday, likely (although I am travelling back that day, so may miss it).

PR #233, opened 2018-11-30
Added basic event routes and event reducers

### LINKED ISSUE #131: Schedule Module API
# Guidelines

Have a short descriptive title <br>
Make sure to click the following:

- [x] Assignees
- [x] Labels
- [x] Milestone

# Full Details

Need to create CRUD routes for Schedule API. Sean has more details on how to elaborate on it

jberns:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #131 

## Description

This adds basic functionality for the event routing and event reducers.


server/models/event.js
```diff
@@ -2,8 +2,32 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const eventSchema = new Schema({
-  title: {
-    type: String
+  master_event_uid:   Number,
+  status:             String,
+  start_date_utc:     Date,
+  end_date_utc:       Date,
+  duration:           Number,
```
seankwilliams: @jberns nice work on this! I'll let Devin offer any feedback he has, my two cents is that this field should be called `duration_in_seconds` instead (or duration_in_minutes) so we know what unit it's using.

jberns: Do you think we even need duration as its own database entry, I can't remember exactly why it was added in the first place?  We can always calculate this from the start/end time of the event.

seankwilliams: @jberns You're right, I think that, out of start date, end date, and duration, we only need two of the three. We can either do without end date or without duration.

jberns: I removed duration as the calendars native input is typically start/end so we don't need to convert between.

dzimmerman902:
Looks good!

PR #248, opened 2018-12-11
Fix Integration Testing

### LINKED ISSUE #234: Backend Testing - Bring it forward
## Description of Issue

At some point during the last few months with all of the improvements, the example backend test I wrote stopped functioning.  This issue will track efforts to fix it and get back on the road to backend testing.

Need to tweak Jest/Supertest/test DB seeding code

msiadak:
## Issue

This issue resolves #234 

## Description

Gets basic HTTP integration testing running.  These tests will fail if you don't have the test DB in your .env file. @dzimmerman902 -- Any suggestions for distributing the .env file?  I can just throw it in Flowdock if you like


dzimmerman902:
@msiadak I went ahead and merged the linting first so it might be the reason it's causing the conflict, that's my bad.

PR #262, opened 2019-01-10
Added new react-scripts, deleted storybook

### LINKED ISSUE #259: Update react-scripts to latest version
## Description of Issue

Currently, the front-end is using react-scripts v1.1.4, which is pretty out of date at this point.  We're also getting pinged by Github for a security vulnerability in an old version webpack-dev-server which is a dependency of react-scripts v1.1.4.

Additionally, upgrading will allow us to use some nice newer tooling seen here:

https://reactjs.org/blog/2018/10/01/create-react-app-v2.html

1. `npm install react-scripts@latest`

2. Test the client, at minimum make sure it still runs without throwing errors.

## Approval?

_Leads Comment Below_

dzimmerman902:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #259 

## Description

Updated package.json to include the updated react-scripts. Problems I ran into was it would not run on newer versions of eslint stuff because of the root (since it was installed there). I added a flag to the .env file to remove the warning and it runs fine. 

@msiadak Could you check to make sure everything is ok? If you remove the SKIP_PREFLIGHT_CHECK from the client .env file, you'll know what I mean.


msiadak:
Should have a chance to take a look this evening.

msiadak:
For reference, here's the error that we were seeing post upgrade:

```
[0] There might be a problem with the project dependency tree.
[0] It is likely not a bug in Create React App, but something you need to fix locally.
[0] 
[0] The react-scripts package provided by Create React App requires a dependency:
[0] 
[0]   "eslint": "5.6.0"
[0] 
[0] Don't try to install it manually: your package manager does it automatically.
[0] However, a different version of eslint was detected higher up in the tree:
[0] 
[0]   <removed>/Comrad/node_modules/eslint (version: 5.12.0) 
[0] 
[0] Manually installing incompatible versions is known to cause hard-to-debug issues.
[0] 
[0] If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
[0] That will permanently disable this message but you might encounter other issues.
[0] 
[0] To fix the dependency tree, try following the steps below in the exact order:
[0] 
[0]   1. Delete package-lock.json (not package.json!) and/or yarn.lock in your project folder.
[0]   2. Delete node_modules in your project folder.
[0]   3. Remove "eslint" from dependencies and/or devDependencies in the package.json file in your project folder.
[0]   4. Run npm install or yarn, depending on the package manager you use.
[0] 
[0] In most cases, this should be enough to fix the problem.
[0] If this has not helped, there are a few other things you can try:
[0] 
[0]   5. If you used npm, install yarn (http://yarnpkg.com/) and repeat the above steps with it instead.
[0]      This may help because npm has known issues with package hoisting which may get resolved in future versions.
[0] 
[0]   6. Check if <removed>/Comrad/node_modules/eslint is outside your project directory.
[0]      For example, you might have accidentally installed something in your home folder.
[0] 
[0]   7. Try running npm ls eslint in your project folder.
[0]      This will tell you which other package (apart from the expected react-scripts) installed eslint.
[0] 
[0] If nothing else helps, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
[0] That would permanently disable this preflight check in case you want to proceed anyway.
[0] 
[0] P.S. We know this message is long but please read the steps above :-) We hope you find them helpful!
[0] 
[0] react-scripts start exited with code 1
```

msiadak:
Yes, no warning for me.  You'll need to delete node_modules and `npm install` in the project root.

dzimmerman902:
Got it, I'll run it myself and if it's good I'll push it to development.

PR #328, opened 2019-02-21
Removed constructor and changed methods to arrow functions. Removed s…

dzimmerman902:
…ome code not needed.

## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #[Insert issue number]

## Description

_Description of pull request_


dzimmerman902:
@seankwilliams Started with restructuring the state and methods. I prefer this style because it's a lot cleaner. When you make the methods arrow functions, it automatically binds *this* so no need to write extra code. I haven't found an article that says which way is better. This way is for sure cleaner.

seankwilliams:
@dzimmerman902 that's great to know, thanks for reviewing! I've made a note to do that in future React work. You can merge this in whenever you're ready.

PR #334, opened 2019-02-27
332 update redux

### LINKED ISSUE #332: Update Redux Folder Structure
## Description of Issue

Josh, Mike and I agreed that the redux structure could use an update. Want to create a directory that's easy to navigate and find what you're looking for.

I showed a photo in the flowdock channel and we agreed it's a good route for now. We are going to divide everything into directories but keep like things together in those directories.

## Approval?

_Leads Comment Below_

dzimmerman902:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #332 

## Description

- We agreed as a group that the redux components could be organized in a more fashionable way. I've moved everything from the Actions and Reducers directory into a redux directory. All the different pieces are separated into their own separate directories with their actions, reducer, and types. 
- Updated all the import routes for all the components using redux.
- Cleaned out some unused code to remove some of the warnings.


dzimmerman902:
Feel free to test it out!

PR #357, opened 2019-03-06
Add additional data to the view track page

### LINKED ISSUE #304: View track page - additional data
depends on #290

Add the following information to the View Track page:
- Track duration, in a mm:ss format (so a 185 second track would display "3:05")
- Track number
- Disk number

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #304 

## Description

_Description of pull request_

### DIFF
--- a/client/src/utils/formatters.js
+++ b/client/src/utils/formatters.js
@@ -1,0 +1,17 @@
+/**
+ * Formats a total number of seconds as mm:ss,
+ * for example, would make 100 seconds 1:40
+ * @param {Number} totalSeconds the total number of seconds
+ * @returns {String} a formatted string in an mm:ss format
+ */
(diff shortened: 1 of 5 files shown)


client/src/utils/formatters.js
```diff
@@ -0,0 +1,17 @@
+/**
```
seankwilliams: @dzimmerman902 When you get a chance, can you confirm that this is the right place and naming for a shared function like this?

dzimmerman902: @seankwilliams Yup! That is exactly where it needs to go!

seankwilliams: thanks!

PR #358, opened 2019-03-08
313 dj dropdown/332 base

### LINKED ISSUE #313: Show - DJ Dropdown Component
required by #239
- [ ] DJ Dropdown Component on instance form 
- [ ] Update clicked show when DJ value changes

jberns:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #313 

## Description

This adds a selection component with an autofilled dropdown for changing Hosts on a show.


dzimmerman902:
@jberns Great work on the code! Is there anything in the code that you would recommend reviewing?

jberns:
If you want to check out the input component real quick, I did add 'dirtyOverride' prop so if you set an initial value you can move the label to the correct place while the field is still pristine.

One thing I would like to add but couldn't figure it out so I left it out for now would be on the blur of the downshift field to replace the current value with the previously selected value if no selection is made (default functionality of downshift), but the reduxform api doesn't seem to be straightforward when it comes to the onBlur function.

PR #367, opened 2019-03-20
286 edit artist name

### LINKED ISSUE #286: View Artist page - allow editing the name
depends on #285

There should be an edit icon next to the Aritst name on the View Artist page. When the user clicks the edit icon, the name of the Artist should be replaced with a textbox with the artist name

When the user clicks anywhere else on the page:
- If the user has not modified the artist name, just change the textbox back to the text showing the artist name.
- If the user has modified the artist name, make an API call to the server to update the artist name. The textbox should disappear and be replaced by a label with the artist name, showing the changed artist name. Also, the Last Updated date/time on the page should be updated.

- Please use React Forms (or whatever form library we end up using - check w/Devin and Josh on this) if possible.
- Use the edit icon from Font Awesome for the edit icon (https://fontawesome.com/icons/edit?style=regular)
- If the user tries to update the artist's name to an empty string, they should receive an error on trying to save.
- If the user tries to update the artist's name to the same name as another artist in a database, they should receive an error on trying to save.
- The artist name should have the white space trimmed by the API endpoint that updates the name.

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #286 

## Description

- Moves View Artist page data to Redux store
- Functionality to let the user edit the artist name
- Adds validation to the artist API


client/package.json
```diff
@@ -25,7 +25,7 @@
     "react-scripts": "^2.1.3",
     "react-table": "^6.8.6",
     "redux": "^4.0.0",
-    "redux-form": "^7.4.2",
+    "redux-form": "^8.1.0",
```
seankwilliams: upgrade not necessarily required - I tried upgrading when I couldn't get initialValues to work, but it turned out the Redux Form version was the sisue

client/src/redux/artist/artistTypes.js
```diff
@@ -1,7 +1,3 @@
-export const ARTIST_ADD = 'artist_add';
-export const ARTIST_CLEAR = 'artist_clear';
-export const ARTIST_ERROR = 'artist_error';
+export const ARTIST_ALERT = 'artist_alert';
+export const ARTIST_EDITING_NAME = 'artist_editing_name';
 export const ARTIST_FIND_ONE = 'artist_find_one';
-export const ARTIST_FIND_ALL = 'artist_find_all';
-export const ARTIST_LOADING = 'artist_loading';
-export const ARTIST_UPDATE = 'artist_update';
```
seankwilliams: removed all unused types

server/controllers/artistController.js
```diff
@@ -1,5 +1,34 @@
 const db = require('../models');
 
+const validateArtistData = async (data, id) => {
```
seankwilliams: this function validates artist data for insert and update functions

dzimmerman902:
LSS (Long Story Short) in regards to the changes. I tried to break up the view page into separate components for clean code. Reason for it is most likely we will be building off these components to be adding new fields/features in the future so wanted to keep that option open. 

I noticed you were probably having trouble with the input focus. ReduxForms has a initial prop that I used to updated the focus state. Much cleaner fix. 

I did take out the pencil, checkmark and x icon. Same reason, it'll be easier to work with the components without those in the future. Plus, we do have button components designed specifically to handle forms and onClick events. I do see a future for them, that's what I created a ButtonIcon component, something easy for a newbie to pick up so they can make more feature rich. 

I did remove some of the alert stuff, didn't want to bloat the reducer with extra fields. I have some ideas on how to handle them a lot better, details coming soon. 

All the functionality is still there and works great, will need to work on some of the UI/UX experience using props for a much more smoother feel. 

I will work on the server side stuff once I get my refactor finished. 

@seankwilliams Let me know if you approve these changes and I'll merge into development. Thanks for all of your great work!

seankwilliams:
@dzimmerman902 Thanks for going through this! Breaking everything out into components makes sense. Although, I'm not crazy about the UI for the form now - the "inline editing" feel was cleaner and more modern to me, and I don't think it will look good once the rest of this page is built out. The rest of the page will list an artist's albums, and with the current Edit button, that would make the Edit button by far the most prominent thing on the page, both because of its color and location in between the title and table of albums. Since the artist name will be the only thing that can be edited on this page, a common UX pattern I've seen for that is having the only editable field able to be edited seemingly inline - you click into it, make a change, and then click out. (Actually, with that UX pattern we may even be able to remove the confirm button - for desktop at least - since clicking out of the field processes an update.) The inline editing feel is also more consistent with the behavior on the wireframes for editing track names.

Regarding this note: 

_ReduxForms has a initial prop that I used to updated the focus state._

That's actually not working for me on my end. When I go to http://localhost:3000/library/artist/5be9d11e294ca24bac186154 and click "Edit", the typing cursor is not in the text field for the artist name. Is that working for you?

Looking forward to seeing your ideas on alerts!

dzimmerman902:
@seankwilliams Yes, that makes sense with the inline editing, you are correct that it will have a better flow and match the wireframe. I still believe we should built out those icons as separate components. I'm going to make an issue for them, that way we can stay consistent and be able to use state, props, etc. for more functionality. 

Didn't know you want the cursor to be inside the input field once you click edit. I will see if I can find a simpler way to make that work. Give me till Thursday and I should have most of this stuff work out.

Sound good?

seankwilliams:
@dzimmerman902 Sounds good, thanks for getting that! I'm all for breaking these out as components. When you're working on it, the form should auto-save when you click out of the field too (that's what the onblur and preventing the default event on button click was for). I can see if I can componentize that functionality after you make your changes. 

Out of curiosity, what type of focus were you referring to with initial prop?

dzimmerman902:
My assumption was your were having an issue with the label part of the input component. Unless you tell the label that there is some value in it, the placeholder would overlap the value so it would look funky. Now I have a better idea of what you are thinking.

seankwilliams:
@dzimmerman902 ah, gotcha!

jberns:
BTW, I had previously added that `dirtyOverride` prop which I think does the same thing as your `initial` prop because I was running into the same issue.

dzimmerman902:
@seankwilliams @jberns Cool, so I added the ability to save with the onBlur, it ended up being WAY easier than I thought. Maybe it might help with your stuff Josh? Redux-Forms is pretty cool on how it handles that stuff. 

Right now it redirects as soon as it saves, mostly because I haven't figured out a way to change the edit state. 

Check it out!

dzimmerman902:
Actually, I know how I'm going to make it work, give me till Thursday lol

dzimmerman902:
@seankwilliams BOOOOOYAH! I think I got all the features you wanted. Consolidated down to one component but the Artist Name is a feature rich form.

I created it with inline styles. You'll have to forgive me if I didn't get the styled exactly before, totally forgot but I insist you had your style flavor to make it look the way you want. Everything is organized so it should be some easy updates.

Thanks man!

dzimmerman902:
Forgot to add the onBlur, I think its good! lol

dzimmerman902:
Shit, nevermind, the onBlur is killing me

dzimmerman902:
I hate this stupid onBlur. It will essentially trigger when the component un mounts. I need to think of a solution.

dzimmerman902:
Ok, I figured out what I was doing wrong. onBlur triggers even when clicking buttons. I looked through Sean's code and found how he solved the problem (so simple, not sure why I didn't think of it lol). So I'll definitely be working on that.

seankwilliams:
@dzimmerman902 nice work on all of this! I didn't know `autoFocus` was a property, very cool. I just committed the style changes.

And regarding onBlur: you got it! The only thing we need to do is be sure the onBlur event doesn't trigger when clicking the checkmark or "x" button. It's okay if onBlur triggers when clicking any other button (like the links in the left sidebar)

dzimmerman902:
@seankwilliams Is there a reason you don't want to use CSS grid?

seankwilliams:
@dzimmerman902 I wasn't able to get the bottom of the pencil icon to line up with the bottom of the text using CSS Grid since they're different font sizes. But I'm not the most adept at CSS Grid, so if there's a way to line up those without having to resort to adding extra margin/padding to the icon, I'm all for it!

dzimmerman902:
Sweet! That's what I'm here for lol. I'm only pushing for CSS grid because it's flexibility for creating layouts and easily adaptable for window screen changes. I just don't want to have to bury components with tons of props and classes when CSS grid can take care of that for us.

seankwilliams:
@dzimmerman902 okay great! Let me know how you end up getting the bottom of the text & icon to line up - I can revert some of my styles to as far as I got with CSS Grid & this if that would be helpful too.

dzimmerman902:
@seankwilliams Alright dude, I think I got everything to make it match with the wireframe. I even added validation so they can't leave it blank. Let me know what you think!

seankwilliams:
@dzimmerman902 very nice work, thanks for spending so much time on this! I'm going to merge this in now.

PR #400, opened 2019-04-02
User form update

dzimmerman902:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves N/A

## Description

I updated the backed User Search feature to be more responsive to the front end user choices. Created a new component called Filter which uses redux-form Field component to capture the data. I put all the pieces together which is what we have is the User Search page. 

It's functional and works great, will definitely want the form to submit after clicking the filter buttons. Right now it submits and then updates the value later which is not what I want it to do. Will make an issue for that.

### DIFF
--- a/client/src/components/Filter/Filter.js
+++ b/client/src/components/Filter/Filter.js
@@ -3,33 +3,25 @@ import { Field } from 'redux-form';
 import classnames from 'classnames';
 
 const Radio = props => {
-  console.log(props.input.checked);
-  return (
-    <input
-      className={props.className}
-      type={props.type}
-      checked={'checked'}
-      {...props.input}
-    />
-  );
+  const { input } = props;
+
+  return <input className="filter__radio" type="radio" {...input} />;
 };
 
 class Filter extends Component {
   render() {
-    const { className } = this.props;
-
-    console.log(this.props);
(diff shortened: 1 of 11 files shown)


dzimmerman902:
@seankwilliams @jberns @msiadak 

So after discussion with Sean and from my personal research, I figured that my instruction for Redux was the wrong approach. Redux should be reserved for the global state like user information. Every other case has been 'one off' scenarios and has been making the code very confusing. Plus, lots of new people are not familiar with it so I think it makes people not want to participate.

Redux will definitely be used but when we really need that global state, I definitely see some potential especially for radio.

That being said, I went ahead and updated the User Search Page. Redux-Form definitely comes in handy still with capturing values and getting the different states of the inputs. I created a Filter component that works like a radio button. 

Let me know what you think, I will most likely merge soon just because I've changed the User Model which breaks the current User Table.

seankwilliams:
@dzimmerman902 looking good to me! I'm not exactly sure yet how to do the forms in a non-Redux way yet but I'll cross that bridge when I get to it.

dzimmerman902:
@seankwilliams Thanks for the feedback! I promise it'll be much easier process without using redux for everything lol

PR #409, opened 2019-04-10
392 checkbox changes

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue
Check progress on issue.

## Description

Want to check how the style looks as well as make sure label is passed correctly into Checkbox component as prop. Is currently passed in from parent component's state.


dzimmerman902:
@kmid5280 Hey Kyle! Below are my notes:

1. Unfortunately I wasn't able to test the code, ran it errors on the UserAddPage. I believe you are missing your Imports for most of those components.

2. If you could, go ahead and make CheckBox a class component. Yes, it doesn't use life cycles, state, etc. but we've found it easier to just make everything class in case things need to be changed in the future (I understand if it started as functional, as I've mentioned, things change all the time!).

3. Go ahead and import the scss file into the scss/index.scss file, that's where we handle those. We use the BEM method for naming our classes, here are more details. https://en.bem.info/methodology/

4. Also, don't worry about tweaking the UserAddPage, I believe @msiadak is working on it. If @seankwilliams told you different, we'll definitely discuss and make some clarification. 

5. As for the redux-form part, take a look at the Filter component. That style seems to fit our needs when developing components for Redux-Form.

6. As for it's look and functionality, looks pretty good. We'll be most likely tweaking the looks more as we move forward.

That's all I can think of from a high level. If you have any questions, please don't hesitate to give myself or @seankwilliams  a text!

kmid5280:
@dzimmerman902 Thanks for the feedback. For clarity, if I use redux-form in the checkbox component, i.e. replace the input tag with the Field tag after importing, the parent component (DashboardHomePage) will have to connect to reduxForm, correct? I'm trying this and this seems to be the error it is giving me.

dzimmerman902:
@kmid5280 Unfortunately I still can't test this, keep getting a broken webpage

kmid5280:
> 
> 
> @kmid5280 Unfortunately I still can't test this, keep getting a broken webpage

@dzimmerman902 Okay, try it now. Initially the code in UserAddPage from lines 52 to 116 was commented out. I initially removed the comments and committed that (beginner mistake probably), so I readded the commenting and pushed it again. For future reference though, how can I best undo that first commit and revert it back to the original? Thanks.

kmid5280:
> 
> 
> @kmid5280 Unfortunately I still can't test this, keep getting a broken webpage

Do I have to make a new pull request?

dzimmerman902:
@kmid5280 Hey Kyle! Thanks for the working on this, a couple of things:

1. I wouldn't recommend "commenting out" code that breaks your work, it's bad practice because you could be possible commenting out someone else's work and could potentially hurt the app.
2. As far as undoing commits, I would use google to research that, I can't think of how off the top of my head.
3. No need to create new pull requests, unless the branch is completely messed up and there is no point of merging it to development, we can always work off just the one pull request and make the changes as we go.

Feel free ot aks any additional questions!

kmid5280:
> 
> 
> @kmid5280 Hey Kyle! Thanks for the working on this, a couple of things:
> 
>     1. I wouldn't recommend "commenting out" code that breaks your work, it's bad practice because you could be possible commenting out someone else's work and could potentially hurt the app.
> 
>     2. As far as undoing commits, I would use google to research that, I can't think of how off the top of my head.
> 
>     3. No need to create new pull requests, unless the branch is completely messed up and there is no point of merging it to development, we can always work off just the one pull request and make the changes as we go.
> 
> 
> Feel free ot aks any additional questions!

@dzimmerman902 Thanks for the feedback. What I meant was that if you look at the UserAddPage on the original development branch, the code is already commented out. I removed them in an attempt to test the component, and recently readded them. I'm not sure why I took them out to begin with.

kmid5280:
@seankwilliams Ok, that should do it for the style. Let me know if there are any other issues.

seankwilliams:
@kmid5280 changes are looking good! Only one piece of feedback functionality-wise: when the checkbox is disabled, can you give it `cursor:default` so it doesn't have a hand cursor when hovering over it? Other than that, we'll just wait for an feedback Devin has.

kmid5280:
@seankwilliams Do you want a hand cursor at all? I could just make it use the arrow in all cases for simplicity's sake.

seankwilliams:
@kmid5280 Yes, any time the checkbox is clickable, it should show a hand cursor. So the only time it shouldn't show a hand cursor is when the checkbox is disabled.

seankwilliams:
@kmid5280 actually my bad, I didn't realize the cursor:pointer behavior was being added through CSS. We should go with the default browser behavior, which is a default cursor above the checkbox, regardless of if it's active, disabled, etc.

kmid5280:
@seankwilliams Ok, so by default cursor, do you mean an arrow regardless of the checked, unchecked, disabled, etc. status? Do you want a pointer (hand) cursor in any situation?

seankwilliams:
@kmid5280 By default, I mean whatever the default behavior for a checkbox is, which may not be the same across OS’s so I can’t give a visual description. If you want to see the default behavior, you can put <input type=“checkbox” /> in a blank HTML file. We should not override the default cursor behavior in any situation.

kmid5280:
@seankwilliams I put the 'cursor: default' value in the checkbox container and removed all the other cursor values. I tried that in both Firefox and Chrome and the result was that it stayed with an arrow pointer in all cases.

seankwilliams:
@kmid5280 great, thanks!

kmid5280:
@seankwilliams Ok got it. Give that a try.

dzimmerman902:
@seankwilliams @kmid5280 Gentleman, I have finished my review! A couple of things to note:

1. Added the propTypes for documentation.
2. Moved label element inside of React component, that way programmers don't need to worry about adding it as a wrapper. 
3. Cleaned up the SCSS and reformatted using BEM methodology.
4. Removed constructor and state, not needed.
5. Added a hover class so user knows when box they are hovering.
6. Made some other stylistic changes. Not a fan of the super round corners as well as increase the size for better visibility. 

If you wish to test out the functionality, start up react-styleguidist (npm run styleguide and then go to localhost:6060). I will also add a photo of the things I've done.

Screen Shot 2019-04-24 at 9 46 15 AM


@kmid5280 Thank you so much for your hard work. As I've said before, things change so please don't take these changes personal. Use them to learn, grow and over time things will become more consistent.

@seankwilliams Let me know if the style works for you or if anything needs to be updated. If not, I'll go ahead and merge into development. 

Thanks!

seankwilliams:
@dzimmerman902 great! Styles look good to me.

kmid5280:
@dzimmerman902 Excellent! Glad I was able to help. Thanks for the feedback.

PR #410, opened 2019-04-10
Artist View Page Update

### LINKED ISSUE #403: No Table Display for No Data
## Description of Issue

@seankwilliams 

What do you think about having some kind of message when there is no data? Instead of having an empty table, maybe have some kind of clean component that essentially has a custom message when there is no data. 

I believe the LargeText component does the trick, maybe we can incorporate it?

- [ ] React Component
- [ ] Message Props

## Additional Comments

N/A

dzimmerman902:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #403 

## Description

@seankwilliams OH YEAH! This is a big update! I've put a lot of work into this to find a functional and practical way to handle the data via redux and I think I'm getting closer to the solution. I'll do my best to explain everything here but feel free to ask me questions!

1. I thought back to original React design on how data is grabbed. Essentially the very root component (in our case, any of the pages) would grab data and pass that data down as props to call the children components. I went back to that philosophy but with redux. Once the ArtistViewPage loads, it grabs the artist data and then the individual components would use the redux as well to grab what data they need.

I find this very efficient and keeps the code clean. Gives it focus on what the purpose of that particular page is for.

2. So I was trying to find an efficient way to handle loading and what displays on the page. Plus, I also wanted to have a way to disable Inputs/Buttons when a form is submitted (so users don't accidentally send multiple requests). What I discovered is redux-form has a prop called submitting. 

In order for it to work, you have to return a promise to the submit function. A simple solution was I would return the last dispatch in artistAlbums as well as the action creator inside the form, that's it! So now we pass that submitting prop to the Inputs/Button which will be true when submitting which will disable them both as well as reduce the opacity so the user knows it's being submitted. 

There's also some features on how to handle if the form fails, I can go in more details as a group.

We can definitely use this strategy for a majority of our forms, keep things in control and make sure it doesn't conflict with any loading features within our app. 

3. To piggy back off my loading philosophy, I thought it would be best to combine the finding albums and finding artist.  My thought process is that trying to do both separately might cause some issues down the road and having to manage two loading states might get confusing. Plus I don't think it hurts that if someone is searching for just one artist, having the albums tag along will most likely be beneficial.

4. I also added the LargeText component when there is no Album data, looks pretty good.

Feel free to pull down my branch and test it out for yourself and let me know if there are any changes you feel would be necessary.

Thanks!


dzimmerman902:
@seankwilliams  Oh yeah, I forgot to mention that was part of some stuff I worked on. I like the approach of having an alert field in the reducers. So pretty much you can pass the whole object (via spead operator) and you also have to pass the action artistCloseAlert (or userCloseAlert, etc. which ever you are using). Seems to work great and gives us an ability to customize messages.

seankwilliams:
great @dzimmerman902 ! I like that approach for alerts.

PR #426, opened 2019-04-25
425 show data seeding

### LINKED ISSUE #425: Data migration script for KGNU shows
## Description of Issue

Write a data migration script to import the shows and show instances into Comrad.

This will need to result in a JSON script that can be imported to the Comrad database. The JSON script can be generated in any way since it will only be run up until the time of the migration to MERN stack Comrad.

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #425 

## Description

Seed data with current show data from Comrad


seankwilliams:
@jberns If it's helpful, here's a branch I'm working on that has all of the current show data from Comrad imported with the seed script. If you use it, please be sure to use a local MongoDB instance instead of the mLab database. This PR is currently making some changes to the users model that are not ready for primetime. I'll need to talk to Devin and see the best way to approach this data since there are required fields in the schema that aren't in the old DB.

server/models/v1/show.js
```diff
@@ -2,26 +2,26 @@ const mongoose = require('mongoose');
 const Schema = mongoose.Schema;
 
 const showSchema = new Schema({
-  status: String,
+  status: String, //what are the possible values?
```
jberns: Right now I am assigning active to all new shows.  I was thinking we can do Active and Cancelled/Deleted at a minimum and add other later as necessary.  I can set this up as an enumeration.

server/models/v1/show.js
```diff
@@ -2,26 +2,26 @@ const mongoose = require('mongoose');
...
-    custom: String,
+    custom: Schema.Types.Mixed, // this will be an object that can contain any number of custom properties
   },
 
   show_start_time_utc: Date,
   show_end_time_utc: Date,
 
   master_show_uid: Number,
-  replace_show_date: Date,
+  replace_show_date: Date, //is this necessary, or redundant with show_start_time_utc?
```
jberns: Yes, when we replace a show with a specific instance we will need to match by the master_show_uid, and the specific show to replace by date.

jberns:
Can you show me how to seed this on Monday?

seankwilliams:
@jberns definitely! I should have show instances ready by then too.

server/models/v1/show.js
seankwilliams: @jberns an enumeration would be great! I'll set "active" as the default.

server/models/v1/show.js
seankwilliams: @jberns got it! Does this make it redundant with `show_start_time_utc` then? Right now I'll be setting the same values for both of these.

jberns: For the most part they would be the same.  But if on the instance you want to reschedule it for some reason, you need one of the fields to be the new date/time while still matching with the original to replace it.

seankwilliams: @jberns ah, that makes sense. thank you!

server/controllers/v1/shows/findAll.js
```diff
@@ -1,7 +1,12 @@
 const db = require('../../../models/v1');
+const keys = require('../../../config/keys');
```
seankwilliams: NOTE: the changes to this file will be written by some of Josh's changes later. They are only here now so the app doesn't freeze when navigating to the Show Calendar page.

dzimmerman902:
Looks good!

PR #427, opened 2019-04-26
427 - Update of dates calculation for individual shows without repeat

### LINKED ISSUE #371: Show Calendar - Bug when changing time of single show
When changing the initial time of a show without repeat, the time is incorrect and then the date is set to the current localized date.

When performing the same action on a show that is set to repeat for a single day, time and date are correct when changing the same time.

Fix:  Adjust date calculation of a single show to match that of the single repeat instance show.

jberns:
…d the data return of individual shows when time was updated

## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #371 

## Description

If dates and times were changed in the new show UI, the function to combine the date and time was not being applied to individual shows.  This PR does two things:
1) Adds function to merge the date and time of individual shows.
2) Adds functionality to the show find controller to return either all dates, or a subset based on a start and end date if provided.


seankwilliams:
@jberns looks great, merging now!

PR #432, opened 2019-04-30
#300 Fixed issue where switching sort tabs didn't update the list correctly

### LINKED ISSUE #300: Music Library: filters should filter rows even when there is no search
Steps to reproduce:
1. Go to http://localhost:3000/library
2. Do not enter any value in the search field
3. Click the "Artists" filter

Expected result:
1. There should be data in the table

Actual result:
1. There is no data in the table

Fixing this will require modifying the `/api/library` endpoint to filter by entity types, if it doesn't already.

Brysonmk1984:
## Guidelines

Development <- sortArtistFix (sorry for this name, it's actually a fix for all the different types).

## Issue

fixes #300 

## Description

When you click on the different sorting tabs on the library page (All, Albums, Artists, Tracks), there is now a call to a specific route on the backend that returns only the specific type that the user clicked on.


server/controllers/libraryController.js
```diff
@@ -0,0 +1,457 @@
+const db = require('../models/v1');
```
seankwilliams: I believe you can remove this file: it was split into separate files and moved into the `/server/controllers/v1/library` folder as part of Devin's refactor (which you pulled into your code when synching with the `development` branch)

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
```diff
@@ -106,7 +130,11 @@ class LibrarySearchPage extends Component {
         activeFilter: event.target.getAttribute('value'),
       },
       function() {
-        this.searchLibrary({ q: this.state.searchString });
+        if (this.state.activeFilter === 'all') {
```
seankwilliams: You will want to add one more check here: you should add `|| this.state.searchString.length > 0` to this `if` condition. The `searchLibrary` function should be called whenever a search string has been provided, and `fetchTableDataByType` should only be called when the search string is empty. `fetchTableDataByType` returns all documents, whereas `searchLibrary` returns the documents filtered by data type (if selected) and the search string.

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
```diff
@@ -75,6 +75,30 @@ class LibrarySearchPage extends Component {
       });
   };
 
+  fetchTableDataByType = type => {
+    axios
+      .get(`/v1/library/${type}`)
+      .then(response => {
+        let pageUrls = this.state.pageUrls;
+        if (typeof response.data.nextPage !== 'undefined') {
```
seankwilliams: You can remove this if block since the  API endpoints you created are not returning `nextPage`.

PR #435, opened 2019-05-02
138 host dropdown

### LINKED ISSUE #138: Host Dropdown
We'll need a dropdown-like component. It should:

1. Like a normal dropdown, have a selected value that appears in the box when it's closed.
2. Have a search-like functionality, where you can start typing a name and the dropdown will filter down to the people matching that search string.
3. Support having one additional values at the top of the list: "Hugo Amorim (me)", where the latter value is based on the currently logged in user. (edit 4/30/19: rather than having "None" as an option, that will be a separate dropdown in the calendar filters)
4. If you search and there are no values, should show a message saying "No hosts found, add a new host" where "add a new host" is a link that will open a popup (popup to be build as a separate component)

Tasks to complete this, as of 4/30/19:
- [x] Create a separate API endpoint with minimal permissions to access (should only return on-air names)
- [ ] Cache results locally and add a delay on API calls so performance isn't laggy
- [x] Change the display of the host dropdown so it only shows on-air names
- [x] Make `HostDropdown` a distinct component
- [x] Property for searching active or all hosts
- [ ] By default, display the currently selected user in the dropdown
- [x] Always display an "Add" option
- [ ] Wire up the "Add" option to a modal form
- [ ] Make `HostDropdown` into a form component rather than a form itself
- [ ] Display the "Me" option at the top of the dropdown
- [x] Have the host dropdown open when clicking into the field
- [ ] Come up with a schema for hosts that are combined users (eg, 99 and Barry)
- [ ] Polish CSS on the dropdown
- [ ] Fix the problem where the text is wrong
- [ ] Open the add modal when clicking away from the field if the host does not exist (or a modal prompting the user to add)
- [ ] When the host dropdown has an existing value, select the text by default on focus

seankwilliams:
Depends on https://github.com/codefordenver/Comrad/pull/431/

## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #138 

## Description

Modify the API to:
- Create a new searchHosts endpoint that is accessible to DJ permission level

Modify host dropdown to:
- cache results and throttle API requests to improve performance and prevent overwhelming the server
- give the option to add a host
- add additional default options to the host dropdown ("Me" and "No Host")
- change host dropdown to not use `ReduxForm`, but instead be able to be used as a `Field` component on Redux Form or as a standalone object
- modify the `Input` component so it does not need to be used within a Redux Form component
- make various UI & UX tweaks


dzimmerman902:
Everything looks good!

PR #438, opened 2019-05-05
#282 Dropdown option to the right

### LINKED ISSUE #282: Dropdown component - create an option for menu to open on the right
The wireframe at https://drive.google.com/file/d/10PNXLQ9ypRs_stgtDLx1bG5gZxzlnbzu/view?usp=sharing shows the "Add" dropdown on the Music Library main page with the menu opening on the right side of the button. However, on http://localhost:3000/library/search the menu opens below the button. 

Please add an option to the Dropdown component that will allow the menu to open on the right side of the button. The default option for the component should still be that the menu appears below the dropdown, but you should be able to pass a property into the Dropdown component that makes the menu open on the right.

Please set the property on the Music Library main page and Users main page so that the menu opens on the right of the Add button.

ehirshfield:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #282 

## Description

@seankwilliams can you take a look at this PR? I added two options for the dropdown. One that does not line up with the add button on the top and one that does line up with the add button on the top. You can swap in those options in the LibrarySearchPage.js file to check them out. I added more items to the dropdown when testing this and it seems the `right-top` class may handle more options in the future better than the `right-centered` class. Also, I think my prettier auto-replaced some hidden lines with spaces in the scss file, let me know if that is ok or not. I can turn that off. 

One last question, should I fix the markdown file with this new option as well?


client/src/components/Dropdown/_Dropdown.scss
```diff
@@ -98,4 +99,26 @@
     bottom: 0px;
     margin-bottom: 4.2rem;
   }
-}
\ No newline at end of file
+
+  & .right-centered {
+    margin-left: 8.5rem;
```
seankwilliams: Rather than hardcoding `8.5rem`, can you use `calc(100% + 0.5rem)`? This will be the same effect, but then the margin will dynamically change if the width of the button is different (like if it has different text).

client/src/components/Dropdown/_Dropdown.scss
```diff
@@ -98,4 +99,26 @@
...
+    left: 0px;
+    transform: translateX(-3px);
+
+    &.active {
+      transform: translateY(0);
+    }
+  }
+
+  & .right-top {
+    margin-left: 8.5rem;
```
seankwilliams: Same comment here - can you switch to `calc(100% + 0.5rem)`?

client/src/components/Dropdown/_Dropdown.scss
```diff
@@ -98,4 +99,26 @@
     bottom: 0px;
     margin-bottom: 4.2rem;
   }
-}
\ No newline at end of file
+
+  & .right-centered {
+    margin-left: 8.5rem;
+    top: -1.5rem;
```
seankwilliams: Can you change this `top` property so it's something dynamic so this is always vertically centered regardless of the number of dropdown items? I don't know how to do that without some research, but I think if you search Stack Overflow for "vertically center with absolute positioning" there's a way to do that.

ehirshfield:
@seankwilliams one more question, I swapped in the right-centered class is that ok? And also the markdown file was `Dropdown.md`

client/src/components/Dropdown/_Dropdown.scss
ehirshfield: Had to dust off the CSS chops for this one for sure haha

seankwilliams: nice!

seankwilliams:
@ehirshfield nice, `right-centered` is looking good. If it's working fine, you can leave `right-top` in, maybe we'll end up using that in the future sometime. And please do update the Markdown file! Then I believe we're good to go.

seankwilliams:
@ehirshfield I synched this branch with the `development` branch. Did you get a chance to update the Dropdown.md file? I'll merge this if it's good to go!

PR #446, opened 2019-05-08
437 album refactor

### LINKED ISSUE #437: Album Refactor
## Description of Issue

Looking to do some refactoring starting with Albums to start thinking in modules. Redux actions are starting to get large so want to separate like we've been doing everything else. 

N/A

## Additional Comments

N/A

dzimmerman902:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #437 

## Description

@jberns @seankwilliams @msiadak 

You ready for this???

Here is the breakdown and then some from Monday. I'll do my best to cover it all:

1. Moved all Album APIs into a separate folder for separation and ease of transferring in case if things were to change. 
2. I imported the react-table css in the root component. Considering we are using it for multiple pages, this is easier instead of having to import it multiple times.
3. Updated TableAlbums tracks with the following:
-- Created a separate component for the custom cell.
-- Moved the handleClick function from the Album view page into the Table component.
-- Updated some values with the new redux style.
4. Made some small changes to TableUsers.
5. Updated AlbumViewPage with the following:
-- Updated with new redux style.
-- Removed handleClick function.
-- Created renderLastUpdated method inside the class instead of calculating inside the render.
-- Change some styling and added text for cleaner look.
6. Updated TrackViewPage, API was going to wrong route
7. Converted the album redux into a new style to help make our process more module base (see me if confusing).

That's the gist of it, I'm probably going to merge soon if you haven't had time to read this which is alright.

Later dudes!


seankwilliams:
nice work @dzimmerman902 !

PR #470, opened 2019-06-07
333 show list

### LINKED ISSUE #333: Show Builder: Show List
The Show Builder icon on the left menu should lead to a page that lists the shows on the current day. There should be an option to change the day of the shows being displayed.

* At the top of the page, have "Display shows for [date selector]", where [date selector] is an input field with a date selector widget. The default value should be today.
  * On updating this, the list of shows should appear immediately below
* There should be a heading on the page that displays the selected date in a format like "Wednesday, February 27"
* The shows for the day should be listed. Each should have the following values:
  * Start time and end time
  * Show name, which is linked to the Show Builder page for that specific show
  * Host name, or "No host" if there has not been a host selected

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #333 

## Description

Changes shows API to:
- sort by show time
- return an array instead of an object with master_time_id keys

Creates the Show List page for Show Builder


client/src/redux/show/showReducer.js
```diff
@@ -55,7 +62,7 @@ export function showReducer(state = initialState, { type, payload }) {
       const searchParams = payload.params;
       return {
         ...state,
-        data: { ...state.data, ...searchData },
+        data: searchData,
```
seankwilliams: @jberns fyi - I'm changing the reducer for finding shows so that it replaces the existing data instead of appending it. Otherwise, there's a bunch of extra show data if you navigate from the calendar to Show Builder. Lmk if this works for you - if not, we can modify things so there's a way to find shows with appending to the existing shows and a separate way to do it with replacing them.

server/v1/controllers/shows/utils/utils.js
```diff
@@ -36,7 +36,28 @@ async function showList(shows, startDate = null, endDate = null) {
...
+    } else if (
+      new Date(a.show_start_time_utc) === new Date(b.show_start_time_utc)
+    ) {
+      return 0;
+    } else {
+      return -1;
+    }
+  });
+
+  return showsToReturnArray;
```
seankwilliams: @jberns Are you okay with changing this to return an array instead of object? I think an array makes more sense based on what a user of the API would expect, and the object couldn't be sorted.

This code is adding ordering to the Find Shows API. The show calendar still works even though this is sending back an array - though I haven't tested it thoroughly so let me know if there's a specific thing I should look at. With these changes there is no master_time_id coming back from this API, so if that's being used somewhere, just let me know where and I'll get it fixed by adding master_time_id as a property to the objects being returned.

client/src/redux/show/showReducer.js
jberns: I'm fine with this, it just means each refresh will be fresh data. We can also update the calendar to only grab the minimum of a weeks worth.

From a refresh side I don't think it would be as nice with the calendar as each week would have to be refetched at the time of navigation.

Alternatively can we run the SHOW_CLEAR when the calendar unmounts, or the show builder mounts and then you have fresh data in the show builder?

server/v1/controllers/shows/utils/utils.js
jberns: I'm fine with returning this as an array.  It may have to be re-keyed on the front-end but that is easy to do in redux.

Also isn't the master_time_id is still one of the object properties?  It is not like we are deleting the property on the object when we key them.

I think everything works fine from the initial display, but the master_time_id is going to be used for making instances, updating shows, and deleting any type of show without requesting the entire time, instead everything is replaced in redux by the key based off of the single show return.

seankwilliams: @jberns ah you're totally right, `master_time_id` is still coming back! I'll leave this returning as an array.

client/src/redux/show/showReducer.js
seankwilliams: @jberns good idea on SHOW_CLEAR, I've got that set up now and undone the change to the show reducer. So we can just keep the show calendar code as is!

seankwilliams:
@jberns this is synched with development now! Can you review & approve when you get a moment, then I'll merge?

PR #471, opened 2019-06-08
339 playlist schema + traffic schema

### LINKED ISSUE #339: Show Builder - Schema
Build the schema for the Show Builder.

* Each show should be represented in the database by a start and end time. 
  * We will not use any identifier for the show instance, because eventually, we want stations to be able to implement the Show Builder module on its own without needing the rest of Comrad.
* Each show should contain a list of Scratchpad items
  * Scratchpad items can be (1) a track, (2) a comment, or (3) a voice break
    * A track will have an associated Track ID
    * A comment will have an associated description (a text value)
    * A voice break will have no additional associated value
  * The scratchpad items will need to have an ordering associated to them so that the order of items on the scratchpad can be remembered.
  * However, the schema will need to be designed in such a way that scheduled traffic events will appear on the scratchpad. Any changes to the traffic events through the Traffic Calendar should be reflected in the Scratchpad items. For example:
    * Say a show from 1pm-2pm has one traffic event: a Legal ID at 1:30pm.
    * The DJ adds Pink Floyd's Time to the Scratchpad before the Legal ID, and Pink Floyd's Wish You Were Here after the Legal ID.
    * Someone goes in and adds a new traffic event: an announcement for Southwest Chief Comedy Festival at 1:30pm.
    * That person also moves the Legal ID to 1:20pm.
    * Now, the Scratchpad should render in one of these two orders:
      * Pink Floyd's Time, Legal ID, Southwest Chief Comedy Festival, Pink Floyd's Wish You Were Here
      * Legal ID, Pink FLoyd's Time, Southwest Chielf Comedy Festival, Pink Floyd's Wish You Were Here
    * Either of these is fine since the system can't really know what the DJ intended when they added these two tracks to the playlist, but the important thing is that the playlist remembers that Pink Floyd's Time is after 1pm but before 1:30pm and that Pink Floyd's Wish You Were Here is after 1:30pm but before 2:00pm.
  * The system should remember the relative position of events so that music tracks, comments and voice breaks in the Scratchpad can always render in the same order that they were arranged by the DJ.
* The Schema should also remember Saved Items
  * This will be similar to the Scratchpad in that it will have ordering of items
  * This will be different than the Scratchpad because in addition to storing traffic, voice breaks, and comments, Saved Items will also include the traffic events. 
  * DJs will add traffic events to the Saved Items in the order that they complete them in the show, so unlike the Scratchpad Items, Saved Items will not have an associated timeframe with them. The database only needs to know about the items' positioning relative to each other. So, it only needs to know that the ordering is Pink Floyd's Time, Legal ID Southwest Chief Comedy Festival, Pink FLoyd's Wish You Were Here. It does not need to know that Pink Floyd's Time was after 1pm but before 1:20pm, the Legal ID was at 1:20pm, Pink Floyd's Wish You Were Here was after 1:30pm but before 2pm, etc.

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #339 

## Description

- Creates schema for playlists
- Consolidates traffic events into single schema
- Creates a base class for event schema that traffic and shows inherit from
- Renamed some properties in the show schema to have more generic names
- Remove unused models


seankwilliams:
@jberns and @dzimmerman902 can you review when you get a chance? This PR makes some schema changes that I'd like your thoughts on:

**Shows and Traffic**
These entities are very similar, so I've made a new "base event" class that each of these inherit from. As part of that, I've renamed a few properties so they don't have "show" in them. I believe I've renamed all of the corresponding properties thoguhout the code, but Josh, I could use your help checking creating/editing show instances to be sure I haven't messed things up there.

Devin - as for traffic, let me know if this works for you. I've removed the separate announcement, giveaway and feature models and combined all traffic into one model instead. I honestly can't remember what we've talked about in the past with that, but I was thinking since we ran into such a mess with artist/track/album being separate models, it's probably best to leave everything in the same model so it simplifies the querying on the back-end. Let me know if that works for you, and I'm definitely open to any changes.

*Removing some unused models*
I've removed the venue and genre models. Genres were in the schema but never used for v1 of Comrad, and since the music library is a secondary piece of Comrad, I don't think we'll need to support this right now. 

As for venues, those are used in ticket giveaways, and with the way the ticket giveaways are used, we can just leave these as text strings in the database. Lmk if you want some more details on that.

**Playlist model**
And, finally, this adds the model for the playlist. Issue #339 lays out the use cases for the schema and I tried to include comments to explain the different properties. Please let me know if there's anything that's not clear.

server/v1/controllers/shows/utils/utils__mongoose.js
```diff
@@ -162,12 +162,12 @@ function master_time_id(_id, start_time) {
 }
 
 function master_time_id__byShowType(show) {
-  if (show.master_show_uid) {
+  if (show.master_event_uid) {
```
jberns: Needs to be changed to master_event_id (currently uid)

jberns:
@seankwilliams  I don't see any problems with sharing the schema between shows and traffic.  I found one error with the master_event_id not being updated, see my comment, but other than that I think this is good!

jberns:
Also all the shows are showing incorrectly, but I assume that the database has to be reseeded.  New show creation seems to be working however.

server/v1/controllers/shows/utils/utils__mongoose.js
seankwilliams: thanks, got it fixed!

seankwilliams:
Thanks @jberns ! Got that change made. And yep, these changes require reseeding the database (I've just been doing that locally), then everything should show up fine. I'll be sure to reseed when I merge this into development.

Issue #465, opened 2019-05-31
Show find API issues


seankwilliams:
The show API is not returning the correct values in the following instances:

**http://localhost:3000/v1/shows/?startDate="2019-05-23T14:22:09.544Z"&endDate="2019-05-23T23:22:09.544Z"**

Actual Results: `[]`

Expected Results: Should return the events between 8:22am-5:22pm on 5/23. Let's include events that span over that timeframe, in other words: if an event starts before 8:22am but is active at 8:22am (ie, an 8am-9am event), it should be included in the results. An event that goes from 5pm-5:30pm should also be included.

**http://localhost:3000/v1/shows/?startDate="2019-05-23T14:22:09.544Z"&endDate="2019-05-24T14:22:09.544Z"**

Actual Results: Returns the following Afternoon Sound Alternative object:
```
{
  "show_details": {
      "guests": [
          null
      ],
      "title": "Afternoon Sound Alternative",
      "summary": "Diverse and eclectic sounds on the more adventurous side.",
      "description": "<p>Diverse and eclectic sounds on the more adventurous side. Tune in for everything from Free Jazz to Hip Hop to Cumbia.</p>",
      "producer": null,
      "host": null,
      "custom": {
          "record_audio": "1",
          "url": "afternoonsound",
          "source": "KGNU"
      }
  },
  "repeat_rule": {
      "byweekday": [
          "MO",
          "TU",
          "WE",
          "TH",
          "FR"
      ],
      "repeat_start_date": "2011-03-28T18:06:00.000Z",
      "frequency": 2,
      "repeat_end_date": "9999-01-01T07:00:00.000Z"
  },
  "status": "active",
  "exclude_dates": [
      "2016-07-29T18:06:00.000Z"
  ],
  "_id": "5cf12122f583ec16304b140d-1558721160000",
  "show_start_time_utc": "2019-05-24T18:06:00.000Z",
  "show_end_time_utc": "2019-05-24T21:00:00.000Z",
  "is_recurring": true,
  "created_at": "2019-05-31T12:42:10.647Z",
  "updated_at": "2019-05-31T12:42:10.647Z",
  "__v": 0,
  "master_show_uid": "5cf12122f583ec16304b140d",
  "master_time_id": "5cf12122f583ec16304b140d-1558721160000"
}
```

Expected results: Should be the 5/23 show, not the 5/24 show

[...]

PR #472, opened 2019-06-08
372 time selector

### LINKED ISSUE #372: Show Calendar - [Spike] Time Selector
Look into if there are better time selectors than the standard form type="time" input, or modifying the default behavior with redux-form

The current input:
- Can be completely cleared using the delete key or putting an invalid number
- Selection is difficult when trying to change time, can only use arrow keys or re-enter numbers manually.
- Field level validation is not working, however should not be required if the delete on the input is completely disabled.

Examples: 
https://react-component.github.io/time-picker/
https://www.npmjs.com/package/react-times
https://reactjsexample.com/simple-react-time-input-field/

jberns:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #372, #465 

## Description

This pull requests addresses the following items:
1) Added a new date picker component to handle both date and time of the show start and end times
2) Updated the show query to populate the master data instead of using async await later (similar to table join at the start)
3) Updated repeat end date calculation for when UTC was midnight, subtracted 1 minute instead of trying to set the hours/minutes on the day.


client/src/components/DatePicker/index.js
```diff
@@ -1 +1,2 @@
-export { default } from './DatePicker';
+export { default } from './DatePicker__AirBNB';
+export { default as DatePicker__React } from './DatePicker__React';
```
seankwilliams: nice work here @jberns !

PR #509, opened 2019-06-26
293 edittrack

### LINKED ISSUE #293: Edit Track functionality
- Add an "Edit" button to the top of the View Track page.  Place the Edit button in a similar position to the album wireframe at https://drive.google.com/file/d/12yDnwB43KrQdMkztHmvm9EbYYX8PABup/view?usp=sharing showing the location of where the button should be
-- The design of the Edit link does not have to match the wireframe
-- Rather than using the Edit pencil in the wireframe, use the edit icon from Font Awesome: https://fontawesome.com/icons/edit?style=regular
- On the front page of the music library, set up the "Edit" link that appears when clicking on the ellipses for tracks to go to the edit page.
- Clicking the edit button should go to a page that uses Redux Form to let the user edit & save the track. The form should be similar to the one used for adding a track.
-- If possible, consider sharing the code between the Add Track & Edit Track forms.
-- The edit interface for the duration_in_seconds field should have two text boxes, one for "minutes" and one for "seconds", so the user would enter [2]:[33] where the brackets represent text boxes.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #293 

## Description

Currently have the basic form but would like some input on how to set the actions and reducers up.

### DIFF
--- a/client/src/components/forms/FormTrackEdit/FormTrackEdit.js
+++ b/client/src/components/forms/FormTrackEdit/FormTrackEdit.js
@@ -10,15 +10,15 @@ import { bindActionCreators } from 'redux';
 class FormTrackEdit extends Component {
   constructor(props) {
     super(props);
-    this.state = {
-      duration_in_seconds: props.duration_in_seconds,
-      minutes: 0,
-      seconds: 0,
-    };
+    this.state = {};
   }
 
   submit = values => {
-    return;
+    const { trackActions, history, albumActions } = this.props;
+    return trackActions.edit(values, trackData => {
+      albumActions.clear();
+      history.push(`/library/album/${trackData.album}`);
+    });
   };
 
   render() {
@@ -26,7 +26,14 @@ class FormTrackEdit extends Component {
     const { handleSubmit } = props;
 
     return (
-      <form className="form-track-edit">
+      <form
+        className="form-track-edit"
+        onSubmit={handleSubmit(data => {
+          submit({
+            ...data,
+          });
+        })}
+      >
         <Field
           component={Input}
           label="Disk Number"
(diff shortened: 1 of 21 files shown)


client/src/components/forms/FormTrackEdit/FormTrackEdit.js
```diff
@@ -0,0 +1,124 @@
...
+class FormTrackEdit extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {};
+  }
+
+  submit = values => {
+    const { trackActions, history, albumActions } = this.props;
+    return trackActions.edit(values, trackData => {
+      albumActions.clear();
```
seankwilliams: Nice job with this here!

client/src/components/forms/FormTrackEdit/_FormTrackEdit.scss
```diff
@@ -0,0 +1,41 @@
...
+    }
+  
+    .duration-colon {
+      order: 3;
+      width: 10px;
+      text-align: center;
+    }
+  
+    .seconds {
+      order: 4;
```
seankwilliams: fyi - I do not think the `order` property is necessary since this numbering is following the order that's in the HTML. We can have it here, just something to know in case this comes up on coding interviews (I could see someone asking whether it's necessary).

client/src/pages/TrackEditPage/TrackEditPage.js
```diff
@@ -0,0 +1,49 @@
...
+      duration_in_seconds,
+      disk_number,
+      album,
+    } = track.doc;
+    return (
+      <div className="track-edit-page">
+        <Card>
+          <CardBody>
+            <h1>Edit Track</h1>
+            <FormTrackEdit
```
seankwilliams: Rather than passing the track data in as properties, can you change the `FormTrackEdit` component so it gets the track data from Redux?

server/v1/controllers/tracks/update.js
```diff
@@ -1,7 +1,13 @@
 const db = require('../../models');
+const { validateEditData } = require('./utils');
 
 function update(req, res) {
-  db.Track.findOneAndUpdate({ _id: req.params.id }, req.body, { new: true })
+  validateEditData(req.body)
```
seankwilliams: Great work with this here!

PR #516, opened 2019-07-07
295 addalbum

### LINKED ISSUE #295: Add Album
depends on #285

Add an "Add" button to the View Artist page that lets you add an album to the artist. 

- The Add button should use a Dropdown component with text of "Add Album"
- When clicking the Add Album button, the user should go to a page that has a form to add the album, using React Form
- The user should receive a validation error if they try to add an album that has the same name as another album by the same artist (it's okay to have albums that are named the same but are by different artists)
- When the user submits the form successfully, they should be returned to the artist page

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #295

## Description

Implement add album form. Link to form is on artist page. Form currently has two inputs, name and label. When form is submitted, it redirects to the artist page showing the new album. Compilation checkbox not included as that will be handled in a separate issue.

### DIFF
--- a/client/src/api/album/add.js
+++ b/client/src/api/album/add.js
@@ -1,6 +1,6 @@
 import axios from 'axios';
-import { ROOT_ARTISTS_URL } from '../root';
+import { ROOT_ALBUMS_URL } from '../root';
 
 export function add(input) {
-  return axios.post(`${ROOT_ARTISTS_URL}`, input);
+  return axios.post(`${ROOT_ALBUMS_URL}`, input);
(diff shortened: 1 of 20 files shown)


seankwilliams:
@kmid5280 this has a few conflicts now - can you resolve them? Other than that, this is looking good!

kmid5280:
@seankwilliams Ok, should be good to go.

seankwilliams:
thanks @kmid5280 !

PR #518, opened 2019-07-09
303 password

### LINKED ISSUE #303: Change password page
The "Change Password" link in #302 should lead to a page where the user can change their password.

The page should have:
- Password field (text box)
- Confirm Password field (text box)
- Submit button

Clicking the Submit button should change the users password, hide the form and display a message saying "Your password has been changed successfully."

The form should require both fields to have values, and the passwords must match. If not, an error message should be displayed.

Please use the same form library as the rest of the application to build this form.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #303 

## Description

Have the basic form set up so far. Link in nav bar takes user to change password page. Wanted to get feedback on how it is so far and what I need to do moving forward to make sure everything is set up right.


ehirshfield:
@kmid5280 so I actually have a branch at the same point you're at haha, should I close mine?

ehirshfield:
Talked to @seankwilliams I'm moving on!

server/v1/controllers/auth/passwordChange.js
```diff
@@ -0,0 +1,45 @@
...
+
+  if (!passConfirm || !passNew) {
+    return res.status(401).json('Password Confirm and/or Password New Missing');
+  }
+
+  if (passConfirm !== passNew) {
+    return res.status(401).json('Your passwords do not match!');
+  }
+
+  const user = await db.User.findOne({
```
seankwilliams: You don't need to find the user to update the password, all you have to do is `db.User.findOneAndUpdate`. You can also remove the validation check to see if it finds `user` below.

server/v1/controllers/auth/passwordChange.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+
+      const updatedUser = await db.User.findOneAndUpdate(
+        { _id: user._id },
+        {
+          password: hash,
+        },
+        { new: true },
+      );
+      return res.json(updatedUser);
```
seankwilliams: Please remove the password property from this object before returning it: the APIs should not expose a password, even if it's hashed.

kmid5280:
@seankwilliams Should be good to go now. Let me know what you think.

seankwilliams:
@kmid5280 This is looking great! Just two more things and we should be ready to merge:

- [ ] Can you resolve the merge conflict? There's one with `index.scss`
- [ ] The Change Password API is still returning the user's password hash. Can you remove this from the response of the API?

password-returned-from-api

kmid5280:
@seankwilliams Had an issue with git pull. Accidentally pulled from development and merged into the branch. I didn't mean to do this. What do you suggest?

seankwilliams:
@kmid5280 You can revert a commit using `git revert (commit hash)`, and if it's a merge, you can use `git revert (commit hash) -m 1` to undo it. But, looking at the commits & changed files, I am not sure you need to do anything. What's the issue you are trying to fix in this branch: is code missing or is there code in the branch there shouldn't be? Let's look at that before doing any git revert: because if you revert code that you in fact need, that can get confusing.

kmid5280:
@seankwilliams My concern is that if you try to merge the branch, there could be a lot of merge conflicts because of the unrelated files that were pulled from development. I'm also not sure if pulling from development affected any files that are specific to this issue, so I wouldn't want to undo any work that I just did. It may not matter in this case, but I want to be sure there won't be any issues with the merge.

seankwilliams:
@kmid5280 Okay, I see what you mean. A couple things here: first, your branch was probably several commits behind `development`, which means there was a bunch of code in `development` that was not in your branch yet. That's why the unrelated files came in. This is a good thing though: because we want the files in your branch _before_ your branch goes to development. Otherwise, we wouldn't know how your code interacts with the changes other people have made until after the Change Password code goes into the `development` branch. If the combination of your changes + the other changes that have happened in `development` do break something, the main branch would be broken which would cause issues throughout the team. So we do want what you did here: all the latest changes are now in your branch, so we can be sure the Change Password changes still work before committing them.

If anyone was changing the same files you were in a way that would break what you were doing, you most likely would have seen merge conflicts. I say "most likely" because you won't always see merge conflicts in that case, but you probably do around 90% of the time. As long as you test your code after the merge, you can be sure that your changes weren't affected.

And, there are no longer any merge conflicts in this branch. You can see that in the pull request conversation in GitHub:

no-conflicts

Let me know if that makes sense and if you've got any further questions about it. I will also run through another test on this code and merge it if it looks good!

seankwilliams:
@kmid5280 this looks good, I'm merging now!

Issue #476, opened 2019-06-08
Delete tracks from Album page


seankwilliams:
Depends on #281 and #283

Set up an ellipsis similar to the one on the Music Library page. It should be in the right side of each row of the track table and clicking it should open a dropdown with "Edit" and "Delete" options. Clicking "Edit" won't do anything yet, but clicking "Delete" should open the same Delete dialog as the Music Library page.

PR #520, opened 2019-07-10
Alerts in Music Library

### LINKED ISSUE #474: Error when displaying Alert component
## Description of Issue

Steps to replicate:
1. Go to http://localhost:3000/library/artist/5be9d30e294ca24bac27eab2
2. Change the artist name to "The Beatles", which will return an error since it has the same name as another artist in the database.
3. You'll get the following error:

Screen Shot 2019-06-08 at 12 35 07 PM

### LINKED ISSUE #489: Need to set up alerts on some music library actions
There are no alerts for errors that happen when adding tracks, deleting albums, and deleting artists. These need to be added.

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #489 
This issue resolves #474 

## Description

- Set up alerts throughout the library. 
- Fixed bug where React Router properties weren't set in Alert sometimes. 
- Fixed problems where form callbacks weren't working properly in some cases. 
- Fixed problems where going directly to some edit pages in the library caused errors. 
- Wired up Edit links on main library page. 
- Set up loading animations throughout music library.

### DIFF
--- a/client/src/api/artist/index.js
+++ b/client/src/api/artist/index.js
@@ -2,10 +2,12 @@ import { add } from './add';
 import { findAlbums } from './findAlbums';
 import { findOne } from './findOne';
 import { remove } from './remove';
+import { update } from './update';
 
 export const artistAPI = {
   add,
   findAlbums,
   findOne,
   remove,
+  update,
 };
--- a/client/src/api/artist/update.js
+++ b/client/src/api/artist/update.js
@@ -1,6 +1,6 @@
 import axios from 'axios';
 import { ROOT_ARTISTS_URL } from '../root';
 
-export function findOne(artistId, propsToUpdate) {
+export function update(artistId, propsToUpdate) {
   return axios.put(`${ROOT_ARTISTS_URL}/${artistId}`, propsToUpdate);
 }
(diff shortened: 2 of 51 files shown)


seankwilliams:
@dzimmerman902 this PR wires up the alerts in the music library! It also adds Redux actions for `alertActions.show` and `alertActions.hide` that can be used when showing/hiding alerts when `dispatch` isn't available. I'm mostly using these to show alerts when the user is redirected after a save operation, and then the alert is shown on the page the user is redirected to. So, for example, after adding an artist, you're redirected to the new artist's page and see an alert saying the artist was created successfully.

I decided against showing the error message in the modal window: the use case was just so narrow for the music library and I think the error messages at the top of the page are just as effective. I just had to add code to automatically close the modal when an error happens.

seankwilliams:
@ehirshfield fyi, if you work on https://github.com/codefordenver/Comrad/issues/476, you'll want to branch off of this branch if the PR isn't merged yet

seankwilliams:
@kmid5280 I made a few changes to the music library forms you had done. The PR looks like like a lot of code, but I wanted to mention two changes to you:

- Ensure the data is loaded if you're navigating directly to a URL for an edit/add page. Previously, the pages were making an assumption that the Redux data for artist/album/track was already loaded, so if you went to a page like Edit Album and refreshed the page, the data wouldn't be there.
- Move the submit callback logic out of the Form components and into the Page. That way, if we reuse the forms anywhere, they don't necessarily have to redirect in the same way the Edit/Add pages are currently. For example, if we put the form in a modal on another page, we can close the modal window on success rather than redirecting the user.

Let me know if you've got questions on any of these! We'll want to put these same changes on the Change Password form: so if you're not sure how to set that up, I'll work with you on it when we do the PR review.

PR #522, opened 2019-07-15
476 delete tracks album page

### LINKED ISSUE #476: Delete tracks from Album page

ehirshfield:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #476 

## Description

Adds an ellipses to the right at the end of the album track row similar to the one on the Music Library page. Edit button leads to edit page. Delete opens a modal to delete.

### DIFF
--- a/client/src/components/tables/TableAlbumTracks/TableAlbumTracks.js
+++ b/client/src/components/tables/TableAlbumTracks/TableAlbumTracks.js
@@ -21,15 +21,10 @@ const CellDuration = ({ value }) =>
 class TableAlbumTracks extends Component {
   state = {
     deleteModal: false,
-    deleteSuccessModal: false,
   };
 
   closeDeleteModal = () => {
     this.setState({ deleteModal: false });
-  };
-
-  closeDeleteSuccessModal = () => {
-    this.setState({ deleteSuccessModal: false });
   };
 
   deleteTrack = id => {
@@ -44,13 +39,11 @@ class TableAlbumTracks extends Component {
   deleteSuccess = entity => {
     this.closeDeleteModal();
     this.props.alertActions.hide();
-    this.setState(
-      {
-        deleteSuccessModal: entity.data,
-      },
-      function() {
-        this.props.handleTrackRefresh();
-      },
+    this.props.handleTrackRefresh();
+    this.props.alertActions.show(
+      'success',
+      'Success',
+      `${entity.data.name} was successfully deleted`,
     );
   };
 
@@ -86,7 +79,7 @@ class TableAlbumTracks extends Component {
 
   render() {
     const { handleRowClick, props, state } = this;
-    const { deleteModal, deleteSuccessModal } = state;
+    const { deleteModal } = state;
     const { albumState } = props;
     const { tracks } = albumState.doc;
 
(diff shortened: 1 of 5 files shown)


ehirshfield:
Hey @seankwilliams the delete modal is feeling like it should become a component itself, thoughts?

ehirshfield:
Also I just need to figure out how to reload the table after delete and this PR is ready for review.

seankwilliams:
@ehirshfield Nice work so far, this is looking great!

A couple comments:
- [ ] Even though the issue says the Edit link won't do anything yet, we actually have that page built now. Can you make it so, when you click Edit, you go to the same page you go to when you click "Edit" in the dropdown for a track on the main library page?
- [ ] You're right in that the delete modal should be a component - do you feel comfortable taking that on as part of this task? If not we can make it a separate task, and let me know if you want me to walk through what you'd do to componentize that piece.
- [ ] To reload the track table after a track is deleted, you'll want to call the Redux action that gets the data for the View Album page again. It's `albumActions.findOne(id);`, and you can see it used in the `componentDidMount` function of `AlbumViewPage`.

ehirshfield:
@seankwilliams let's make the modal split out another task so I don't keep this one waiting. I can move to it right after! Also the findOne method from album tracks doesn't seem to hot reload like the library page, mind taking a look? I'll take another crack at it tomorrow night too but from tinkering around with that method on the albumViewPage as well as on the TableAlbumTrack does not work. I was hoping to not have to do it like the LibrarySearchPage haha but we'll see

seankwilliams:
@ehirshfield sounds good, I've made another issue at https://github.com/codefordenver/Comrad/issues/538 and assigned it to you! You totally have the right idea with findOne: the problem is that `albumActions` is not defined in the `handleTrackRefresh` function. I don't know why it doesn't throw an error, but I've ran into that problem in some other places too and I wish it did (it would make troubleshooting so much easier!)

ehirshfield:
Wow what a simple fix, thanks @seankwilliams! Should be good to review, there is one thing I want to note to do later and it's to make sure the edit/delete dropdown comes to the forefront when opened near the bottom of the list. Looks like a z-index issue or something? Anyway, let me know if I should address that here!

seankwilliams:
@ehirshfield This is looking great! One last thing: now, the delete success modal isn't showing for some reason. I'm not quite sure why but we should get rid of that modal anyway and replace it with the built-in alert system. Can you remove the delete success modal and use `alertActions.show` to display a success message when a track is deleted instead?

As for the issue at the bottom of the table, I see what you mean. I'll log an issue for that later this week, as I think that problem is related another one we're having where scroll bars are appearing when they shouldn't within some of our tables.

ehirshfield:
@seankwilliams should be good to re-review! Also, check out my fix in this PR for the react-table. I saw some double scroll bars and looked into it while working on this, but feel free to tell me to take it out if it causes adverse side effects. I looked around and think it's ok. At least on Mac haha

seankwilliams:
@ehirshfield Nice work, this looks great! Merging now. And thanks for looking at the scroll bar issue: I'll check out that scroll bar fix either tonight or tomorrow!

PR #525, opened 2019-07-17
500 addbutton

### LINKED ISSUE #500: "Add" button on calendar views
The Show Calendar and Traffic Calendar should have an "Add" button at the top of the page that can be used to add events. Even though events can be added by clicking on a blank space on the calendar, an "Add" button will make the UX clearer and also provide an easy way to add events when the calendar is crowded.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #500

## Description

Implement add button that user can click on to add new show to calendar.


kmid5280:
Note: When I add a new show, the new show appears on the calendar in this sort of random spot instead of on the time slot that was entered. This happens whether I use the add button or an empty time slot.
calendar-time-slot

seankwilliams:
@kmid5280 Looking great so far! I noticed these two things, can you take a look?

- [ ] Can you display a "+" icon on the Add button? It should look liek the Add button that's on the Library page.
- [ ] Can you add a little margin below the Add button so it's not pressed against the mini-calendar?

Also, I looked a little closer at the problem you mentioned where new shows are appearing in a random spot, but when I look closer at your screenshot, it does look like the shows are appearing where they're supposed to. "ddd" is at 10am, the other new show is at 9am. Are these appearing at different spots then the values you entered? I haven't been able to replicate, so if you can get the issue again, can you write out the steps to reproduce?

The shows appearing in different colors on the calendar is the expected: I'm not sure what the colors mean, but Josh has them set up to mean different things to help with development. In production, they'll all be the same color.

kmid5280:
@seankwilliams Ok give that a try. It should look okay now, but I had some trouble putting the scss style together. For instance using "&-text" didn't work, I had to use "& .calendar-add-button-text". Any thoughts on how to put that together a little better?

seankwilliams:
@kmid5280 That's looking a lot better! Now the main calendar has been pushed down, can you adjust things so it's still at the top of the area? 

calendar-space

Regarding `&-text`, that worked for me using this code:

text-scss

If you want to commit the code you were having trouble with, I can take a look.

kmid5280:
@seankwilliams That's strange, I tried using &-text, but it didn't work. I just tried it again and it did work. Not sure what changed. Anyway take a look now, should be good.

client/src/pages/CalendarHomePage/CalendarHomePage.js
```diff
@@ -64,6 +74,7 @@ class CalendarHomePage extends Component {
       onlyDisplayShowsWithNoHost,
       selectedDate,
     } = this.state;
+    console.log(this.props);
```
seankwilliams: @kmid5280 can you remove this `console.log`?

client/src/components/Shows/NewShow/Form.js
```diff
@@ -88,15 +88,15 @@ class NewShowForm extends Component {
               )}
 
               <Field
-                className="grid-span--2"
+                className="grid-span--2, form-summary"
```
seankwilliams: @kmid5280 can you undo this and the `form-desc` class you're adding below? (That's what's changing the fields from being stacked to appearing left/right of each other, right?)

seankwilliams:
@kmid5280 this is looking fantastic! I left a few comments in the PR that are some minor things, and then I'll do a full review tomorrow or over the weekend and merge this in. Thanks for the great work on this!

kmid5280:
@seankwilliams Ok, should be good to go now.

client/src/pages/CalendarHomePage/CalendarHomePage.js
```diff
@@ -33,6 +36,13 @@ class CalendarHomePage extends Component {
     };
   }
 
+  showNewShowModal = show => {
```
seankwilliams: @kmid5280 is the `show` property necessary here? I don't see it being passed in from where `showNewShowModal` is called.

kmid5280: @seankwilliams In the initialValues in Form.js, it checks to see if there's an existing show to determine whether to populate the date fields. If I remove the show param from the showNewShowModal function, then that ceases to work and it ends up populating the fields with the current date. The choice I think would be to either leave it as is or find a different way to show an empty field if the add button is clicked rather than a slot on the calendar.

seankwilliams: @kmid5280 Okay, I see. Right now, `show` in this function is set to a `MouseEvent` because it's firing from a click event. So while it's working, we'll want to change this a bit so it's not setting a bunch of values for `MouseEvent` in the `show.selected` state when that is expecting to receive a show. Otherwise, someone might try to rely on the data in `show.selected` in the future only to find that it's in an unexpected format.

To do this, rather than calling `selectShow`, you'll want to add a new Redux action that clears the selected show and sets the `show.selected` value to `{}`. And, I don't think it's necessary to pass `show` into `setModalVisibility` but I didn't go through the whole modal thoroughly so I'm not positive. If it is necessary, you'll want to pass in something like `{}` instead so it's passing in an empty show rather than a `MouseEvent`.

Let me know if that makes sense!

kmid5280: > 
> 
> @kmid5280 Okay, I see. Right now, `show` in this function is set to a `MouseEvent` because it's firing from a click event. So while it's working, we'll want to change this a bit so it's not setting a bunch of values for `MouseEvent` in the `show.selected` state when that is expecting to receive a show. Otherwise, someone might try to rely on the data in `show.selected` in the future only to find that it's in an unexpected format.
> 
> To do this, rather than calling `selectShow`, you'll want to add a new Redux action that clears the selected show and sets the `show.selected` value to `{}`. And, I don't think it's necessary to pass `show` into `setModalVisibility` but I didn't go through the whole modal thoroughly so I'm not positive. If it is necessary, you'll want to pass in something like `{}` instead so it's passing in an empty show rather than a `MouseEvent`.
> 
> Let me know if that makes sense!

You're suggesting to implement a new action that clears the selected show. However by clicking the Add button, we are assuming that there is no show data to begin with. Is implementing this action supposed to be a defensive way of making sure there is no existing show data, or are we always going to have show data no matter the situation?

kmid5280: @seankwilliams

seankwilliams: @kmid5280 It's a defensive way to be sure that if there is show data, it gets cleared. There will probably be some instances where show data isn't cleared from Redux so this will be sure the show data gets cleared when that happens.

seankwilliams:
@kmid5280 the work you have for the `clearShow` action looks good, so let's just modify the `showNewShowModal` so it:
- doesn't call `selectShow`
- calls the `clearShow` action instead
- doesn't pass a `MouseEvent` to the `setModalVisibility` function

Does all that make sense? I can explain the `MouseEvent` thing some more if that isn't clear.

kmid5280:
@seankwilliams I have the action and the reducer set up, but I don't think it's correct. When clicking the add button I get an error that says "TypeError: selectedShow is undefined." The clearShow action calls the SHOW_CLEAR type which is already in the reducers, but dispatches a payload of {}. Is this the right idea or do I need to do it differently, i.e. set the value of show.selected through the reducer?

Also if you could go over the mouse event thing in a little more detail that would be great.

seankwilliams:
@kmid5280 For `MouseEvent`, can you console.log the value of the `show` parameter being passed to the `showNewShowModal` function here? Then I'll explain a few things with those values.

As for the `TypeError: selectedShow is undefined` error, which file is that being thrown in? It sounds like the New Show modal might not be set up to handle when a selected show is not set, so you'll want to change the modal to account for when a show is not set at all.

seankwilliams:
@kmid5280 This looks good, nice work! I committed some changes that fix a merge conflict that was resolved incorrectly: some of the show types in `development` have changed and this PR was adding show types back in that were no longer in use. Merging now!

PR #535, opened 2019-07-20
529 scroll

### LINKED ISSUE #529: Library: Auto-scroll user to top of page after clicking "Yes" on delete modal
## Description of Issue

In the library, if you scroll down and delete a record, after you click "Yes", you can sometimes get validation errors that appear. But, they're at the top of the page, so the user can't see them. To fix that, let's auto-scroll the user to the top of the page every time after a user clicks "Yes" on the Library delete modal.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #529 

## Description

Implement auto-scroll to top of page when user selects item on librarysearchpage, clicks delete, then clicks yes. Used window.toScroll() in deleteSuccess and deleteFailure functions.

### DIFF
--- a/client/src/api/genre/findAll.js
+++ b/client/src/api/genre/findAll.js
@@ -1,0 +1,6 @@
+import axios from 'axios';
+import { ROOT_GENRE_URL } from '../root';
+
+export function findAll() {
+  return axios.get(`${ROOT_GENRE_URL}`);
+}
--- a/client/src/api/genre/index.js
+++ b/client/src/api/genre/index.js
@@ -1,0 +1,5 @@
+import findAll from './findAll';
+
+export const genreAPI = {
+  findAll,
+};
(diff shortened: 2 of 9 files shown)


seankwilliams:
@kmid5280 I'm seeing a few things here so far, can you take a look at these?

- [ ] The Genre dropdown does not appear on the Add Album page (it's there on Edit, but not on Add)
- [ ] I'm getting an error when trying to update an album on the Edit Album page:
edit-album

- [ ] The genre is not appearing on the View Album page (this is using my local database so you may not be able to test with the same album in development):
view-album-page

kmid5280:
@seankwilliams To pass the genres into the album edit form, the findAll function is called in the previous component, and then is passed in as props to the form itself. The album add page is accessed from a different component, i.e. the library search page, so where do you suggest I run the findAll function for the album add page? Should I also run it from the library search page or within the add album page itself?

seankwilliams:
@kmid5280 I realize I left these comments in the wrong PR: moving over to #543

kmid5280:
@seankwilliams Closing PR, moving issue to PR #556

PR #537, opened 2019-07-23
526 loading

### LINKED ISSUE #526: Loading component on Dashboard appears three times, stacked on top of each other
## Description of Issue

There are three `Loading` components on the Dashboard, and they all appear right on top of each other. We should modify the `Loading` component so it can be shown just in a particular section rather than across the whole screen: that way, each of the three `Loading` components shows within the section of the Dashboard that is being loaded.

Add a new property to the `Loading` component that lets it display as a statically positioned element rather than an absolutely positioned element. Change the `Loading` components on the Dashboard to use that new property.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #526 

## Description

### DIFF
--- a/client/src/components/Loading/_Loading.scss
+++ b/client/src/components/Loading/_Loading.scss
@@ -10,7 +10,7 @@
   text-align: center;
 }
 
-.__gif {
+.loading__gif {
   height: 10rem;
   width: 10rem;
 }
--- a/client/src/components/Loading/_Loading.scss
+++ b/client/src/components/Loading/_Loading.scss
@@ -1,13 +1,16 @@
-.loading {
-
+.absolute {
   position: absolute;
   top: 50%;  
   left: 50%; 
-
   transform: translate(-50%, -50%);
+}
 
-  &__gif {
-    height: 10rem;
-    width: 10rem;
-  }
+.static {
+  position: static;
+  text-align: center;
+}
+
+.__gif {
+  height: 10rem;
+  width: 10rem;
 }
--- a/client/src/components/Loading/_Loading.scss
+++ b/client/src/components/Loading/_Loading.scss
@@ -1,11 +1,11 @@
-.absolute {
+.loading--absolute {
   position: absolute;
   top: 50%;  
   left: 50%; 
   transform: translate(-50%, -50%);
 }
 
-.static {
+.loading--static {
   position: static;
   text-align: center;
 }
--- a/client/src/pages/DashboardHomePage/DashboardHomePage.js
+++ b/client/src/pages/DashboardHomePage/DashboardHomePage.js
@@ -52,11 +52,9 @@ export default class DashboardHomePage extends Component {
 
     return (
       <div className="dhp">
-        {/*loadingOnAirData && <Loading />*/}
-
         <Card className="dhp__section-1">
           <CardBody>
-            <div style={{ position: 'relative' }} className="upcoming-shows">
+            <div className="upcoming-shows">
               <h2>My Upcoming Shows</h2>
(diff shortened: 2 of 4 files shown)


kmid5280:
@seankwilliams Give this a try. I ended up not using the classnames library, and for some reason the scss file wasn't recognizing the & pseudo selector. The conditional in the loading component should set the default className to 'absolute' but can change to 'static' if that value is passed in.

client/src/components/Loading/_Loading.scss
```diff
@@ -1,13 +1,16 @@
-.loading {
-
+.absolute {
```
seankwilliams: These CSS changes are the right idea, but there's two things we should change. First, the class names `absolute` and `static` are generic and risk conflicting with other classes we might use in the future. In general you want to name your CSS so that it's tough to have duplicate names. Second, we'll want to implement BEM naming here. If you're not familiar with it, please review at http://getbem.com/. Using BEM naming, we can come up with some non-generic names that explain these elements and the hierarchy between them: for example, you could use `.loading--absolute`, `.loading--static`, and `.loading__gif`.

client/src/pages/DashboardHomePage/DashboardHomePage.js
```diff
@@ -52,19 +52,20 @@ export default class DashboardHomePage extends Component {
...
             <div className="upcoming-shows">
               <h2>My Upcoming Shows</h2>
-
-              <ShowListForUser
-                maxItems="3"
-                startDate={today}
-                endDate={todayPlus3Months}
-                noItemsText="You have no upcoming shows in the next three months."
-              />
+              {!loadingOnAirData ? (
```
seankwilliams: Since `loadingOnAirData` is not related to the data in the `ShowListForUser` component, the component should load immediately rather than waiting for `loadingOnAirData` to be true. You also don't need to add a `Loading` component here, because there's already one inside the `ShowListForUser` component. You'll want to go into the `ShowListForUser` component and change the `Loading` component there so it uses the new style you've set up.

client/src/pages/DashboardHomePage/DashboardHomePage.js
```diff
@@ -73,81 +74,101 @@ export default class DashboardHomePage extends Component {
...
+                          {moment(currentlyOnAir.start_time_utc).format('LT')} -{' '}
+                          {moment(currentlyOnAir.end_time_utc).format('LT')}
+                        </div>
+                      </>
+                    ) : (
+                      <LargeText align="center">No Show On Air</LargeText>
+                    )}
+                  </>
+                ) : (
+                  <div>
```
seankwilliams: Is it necessary to wrap this in a `div`? If not, please remove the div wrapper.

kmid5280:
@seankwilliams I made the changes. Regarding the scss, is that more in line with what you were looking for?

seankwilliams:
@kmid5280 For the scss, that's closer, but still not it. There shouldn't be a class called `__gif`: in BEM, there's no situation where a class would start with a double underscore. It should be `loading__gif` instead, since it's within the `loading` element.

kmid5280:
@seankwilliams I made some changes. Something like that you mean?

seankwilliams:
@kmid5280 Yep, that's much better! Next, can you get the comment I left above related to `loadingOnAirData` and `ShowListForUser`? After that I believe this will be ready for review & merging.

kmid5280:
@seankwilliams That ought to do it. Try it now.

seankwilliams:
Nice work @kmid5280 , this looks great. Merging now!

PR #543, opened 2019-07-25
502 genre

### LINKED ISSUE #502: Add Genre to music library
Albums will need to have genres for use in reporting. For example, music directors will want to pull the most popular tracks played in a given month for a particular genre. Add a Genre dropdown to the Edit Album and Add Album pages.

- [ ] Set up an API endpoint to get all genres from the database (they are in the `genre` model)
- [ ] Set up a new set of Redux action, reducer and type for genres. Model this off of the folder and file structure that the `album` redux function uses.
- [ ] Set up a Redux action to retrieve the genres from the new API endpoint and place the  genre list in the Redux store
- [ ] Modify the Edit Album and View Album page so that, if the genre list is not loaded into the Redux store on page load, the page calls the Redux action to retrieve genres
- [ ] Add a dropdown to the Edit Album and View Album pages with the options for genre. This field should be optional.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #502 

## Description


kmid5280:
@seankwilliams Let me know your thoughts so far.

seankwilliams:
@kmid5280 looks good at quick glance going through, I'll pull the code and test it out once you have things fully working!

seankwilliams:
Adding this comment that I accidentally put in #535  here:

@kmid5280 I'm seeing a few things here so far, can you take a look at these?

- [ ] The Genre dropdown does not appear on the Add Album page (it's there on Edit, but not on Add)
- [ ] I'm getting an error when trying to update an album on the Edit Album page:
edit-album

- [ ] The genre is not appearing on the View Album page (this is using my local database so you may not be able to test with the same album in development):
view-album-page

seankwilliams:
@kmid5280 in response to this:

_To pass the genres into the album edit form, the findAll function is called in the previous component, and then is passed in as props to the form itself. The album add page is accessed from a different component, i.e. the library search page, so where do you suggest I run the findAll function for the album add page? Should I also run it from the library search page or within the add album page itself?_

I would have the `findAll` function store its results in Redux (I think it is currently but I haven't checked so I might be wrong). Then, any time you need to access those results, use them from Redux's store. If the results aren't set, that's when you should call `findAll` so it populates Redux's store.

kmid5280:
@seankwilliams In that case would you suggest calling it from the dashboard, and then putting it in the store to start with that way? That way it's preloaded and I can access it from anywhere.

seankwilliams:
@kmid5280 I would just put it on the pages it's used so that the code is close to where it's used. That will help when other people are working on the code. Also, it's going to eventually be possible to log in and bypass the dashboard page: for instance, if you're trying to navigate directly to Music Library and need to authenticate, the login screen will send you directly to the Library page (the page you were initially trying to navigate to).

kmid5280:
@seankwilliams Ok, so I have a findAll function being called on the AlbumViewPage, even though the genres are being listed on FormAlbumAdd. Would you suggest moving that function to the form instead of passing it into the form from the album view page?

seankwilliams:
@kmid5280 Yes, that sounds good to me: go ahead and move the findAll function from AlbumViewPage to FormAlbumAdd.

kmid5280:
@seankwilliams I was able to move the findAll action into the album add form and album edit form, and remove it from the album view page. It seems to be working okay as far as I can tell. One issue I'm running into is that if I access the album add form from the add button on the library search page, it will simply display a loading graphic and won't open, but if I open from the artist view page, it will load. Presumably it's expecting an artist to be there. Is that a separate issue that still needs to be fixed?

I haven't yet fixed the error that you mentioned, that you get when updating the album.

kmid5280:
@seankwilliams Also I haven't fixed it to populate the genre on album view page either, that's still on the todo list.

seankwilliams:
@kmid5280 okay great! Don't worry about that error when clicking to Add Album from the main library page: that's a different issue that we'll take care of separately (it's an issue with how the routes are configured).

client/src/components/Select/Select.js
```diff
@@ -30,11 +34,21 @@ class Select extends Component {
...
-          )) || null}
+          {isObj
+            ? Object.keys(selectOptions).map((option, index) => {
+                return (
+                  <option key={index} value={selectOptions[option]._id}>
+                    {selectOptions[option].name}
+                  </option>
+                );
+              })
+            : selectOptions.map((option, index) => {
```
seankwilliams: Nice work with this here!

PR #554, opened 2019-07-31
Created a new Library model that combines artists, albums and tracks …

### LINKED ISSUE #327: Library Search Page Code Refactor
## Description of Issue

Going to go through Sean's code and refactor to match our styles. I will tag him along the way to show what can be updated.

Lots of code

## Approval?

_Leads Comment Below_

### LINKED ISSUE #445: Bug: Music Library table's descending sort order is not working
Steps to replicate:
1. Navigate to http://localhost:3000/library 
2. Click "Updated At" in the table header

Expected behavior:
The records should be sorted with a sort order of updated_at ascending, so the records with the oldest updated_at time should be displaying first.

Actual behavior:
The records with the most recent updated_at time appear first.

This behavior also affects the other columns on the table. It seems that columns are sorting in ascending order regardless of the value passed into the sortDescending parameter in the API.

### LINKED ISSUE #540: Create Paginated Results - Albums
## Description of Issue

With Albums, using the findAll can be painful because it has to pull thousands of documents. 

Tweak the findAll controllers to paginate the results. Have some default values and parameters to change them.

### LINKED ISSUE #541: Create Paginated Results - Tracks
## Description of Issue

With Tracks, using the findAll can be painful because it has to pull thousands of documents. 

Tweak the findAll controllers to paginate the results. Have some default values and parameters to change them.

### LINKED ISSUE #542: Create Paginated Results - Artists
## Description of Issue

With Tracks, using the findAll can be painful because it has to pull thousands of documents.

Tweak the findAll controllers to paginate the results. Have some default values and parameters to change them.

### LINKED ISSUE #553: Music Library: Back-end refactor
Refactor the model for the music library so that the APIs behind the library search isn't so confusing

seankwilliams:
…so that the search queries in the main library page can be simplified

## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #327 
This issue resolves #445 
This issue resolves #553 
This issue resolves #540 
This issue resolves #541 
This issue resolves #542 

## Description

- Combines artists, albums and tracks into one model to simplify the queries that search through all three
- Fix an issue where the descending sort option wasn't working in the library
- Change the main Library page to use the Redux patterns used in the rest of the app


seankwilliams:
@jberns and @dzimmerman902 , can I get your thoughts on this before I move forward? 

Before I bring the Library functionality into Show Builder, I wanted to go back and clean up some things around the Library code. I'm starting to do what we talked about a few months ago: combining the artists, albums and tracks into one model. Except, this will actually get rid of the artists, albums and tracks models completely, which is different than what we previously discussed. 

The reasoning here is that a Mongo text index is used for the search, and to use that, whatever model we are searching has to have the field we are searching - in our case, the `name` field. It can't reference another model, unfortunately. So if we left the artists/albums/tracks models in place, we would need to duplicate the `name` field between those three models and the combined `Library` model. I think it's cleaner if we put everything into one `Library` model: that way, we don't have duplicate data anywhere in the database. And, we already had a `type` field that specified album/artist/track, so we still have that on the `Library` model and can use that for filtering. 

The two big benefits here are the back-end code is way cleaner (no more albumSkip/artistSkip/trackSkip nonsense) and the search API is running faster (it was taking 1.2s before, now it's at 600ms). The downside is developers may (1) get confused when they go looking for a track, artist, or album collection in the database and (2) have to remember to filter by `type` on certain queries.

Do you think this is a good approach? If so, I'll proceed with removing the artist/album/track models and updating all the back-end code.

dzimmerman902:
@seankwilliams I'll have to look deeper when I get a chance. My only question is let's say you need to update the spelling of an artists name. Do you have to update all songs/albums or will it only affect the one document?

seankwilliams:
@dzimmerman902 Thanks! You'll just need to update the artist's record. There's still an `artist` field like there was on albums, it just references another `Library` record now.

seankwilliams:
@kmid5280 , @dzimmerman902 , @jberns This refactors the library so that:

- The whole library uses one `Library` collection rather than separate `Artist`, `Album`, and `Track` collections. This cleans up the complicated logic around the back-end search. The back-end controllers and front-end Redux files have been combined for these.
- Fixes a bug where the sort order wasn't properly sorting in a descending fashion when clicking on headers on the library page
- Refactors the Library search page to follow our Redux conventions (previously, it was doing most of its work within the state of the `LibrarySearchPage` rather than using Redux)
- Changes the genre dropdown to be ordered alphabetically
- Adds missing custom fields (Location + Is Local) to the Add Album page

Let me know if you have any questions! Also, I would review using https://github.com/codefordenver/Comrad/pull/554/files?w=1 so the whitespace differences don't become confusing.

seankwilliams:
And, I forgot to mention: this will require reseeding the database, so let me know if you'd rather I merge & reseed the dev database first before you review. (It'd be easiest to reseed after this is merged since we're against the mlab file size limit, but I could try deleting a bunch of records from the existing collections and creating a limited number of records in the Library collections if you want to test against the dev db before merging this.)

PR #556, opened 2019-08-07
scroll to top of page during deleteFailure and deleteSuccess functions

### LINKED ISSUE #529: Library: Auto-scroll user to top of page after clicking "Yes" on delete modal

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #529 

## Description

### DIFF
--- a/client/src/pages/LibrarySearchPage/LibrarySearchPage.js
+++ b/client/src/pages/LibrarySearchPage/LibrarySearchPage.js
@@ -56,10 +56,12 @@ class LibrarySearchPage extends Component {
   };
 
   deleteFailure = () => {
+    window.scrollTo(0, 0);
     this.closeDeleteModal();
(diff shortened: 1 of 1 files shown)


kmid5280:
@seankwilliams You mentioned that when you went to the library page and deleted an album that has tracks, it wouldn't scroll to the top of the page to show the error. It seems to work okay for me now, so try this and see if this fixes it.

seankwilliams:
@kmid5280 this looks good, merging now!

PR #566, opened 2019-08-13
Changed access control around searchHosts endpoint. Integrated host d…

### LINKED ISSUE #337: Show Builder - Integrate host dropdown
depends on #138
depends on #335

* Bring the host dropdown component from #138 into the Show Builder page. 
* The host dropdown should default to the host currently assigned to the show instance, or should be empty
* Changing the host dropdown should automatically update the host for that show instance in the database

seankwilliams:
…ropdown with Show Builder.

## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #337 

## Description

- Adds host dropdown to Show Builder

### DIFF
--- a/client/src/components/Shows/ViewShow/Form.js
+++ b/client/src/components/Shows/ViewShow/Form.js
@@ -11,19 +11,11 @@ import {
   createInstanceAndEditShow,
   updateShow,
 } from '../../../redux/show';
+import { getShowType } from '../../../utils/shows';
 import { MODAL_EDIT_SHOW, MODAL_EDIT_SERIES } from '../ShowModalController';
 
 import DropdownHost from '../../DropdownHost';
 
-export function getShowType(show) {
-  if (show.is_recurring) {
-    return 'series';
-  } else if (show.master_event_id) {
-    return 'instance';
-  } else {
-    return 'regular';
-  }
-}
 class NewShowForm extends Component {
   componentDidMount() {
     const { setModalVisibility } = this.props;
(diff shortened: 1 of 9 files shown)


client/src/components/Shows/ViewShow/Form.js
```diff
@@ -11,19 +11,11 @@ import {
   createInstanceAndEditShow,
   updateShow,
 } from '../../../redux/show';
+import { getShowType } from '../../../utils/shows';
```
seankwilliams: @jberns This PR moves `getShowType` into a `utils/shows.js` file - does that work for you?

client/src/redux/show/showActions.js
```diff
@@ -133,6 +134,8 @@ export const searchShow = (
   onlyDisplayShowsWithNoHost = false,
 ) => async dispatch => {
   try {
+    dispatch({ type: SHOW_FETCHING });
```
seankwilliams: @jberns Also, does this work for you? I've added this into the `searchShow` action so that `state.show.fetching` is set to `true` when a show search is going on.

server/v1/models/accessControl.js
```diff
@@ -4,18 +4,22 @@ const Schema = mongoose.Schema;
...
   },
 
   action: {
     type: String,
+    required: true,
   },
 
   attributes: {
     type: String,
+    required: true,
```
seankwilliams: @dzimmerman902 fyi - this PR sets the properties in the AccessControl collection to required. Otherwise, I was finding there would sometimes be issues in the CLI that would create empty records, and those cause the access control library to break.

PR #567, opened 2019-08-13
import withRouter, wrap connect function with withRouter

### LINKED ISSUE #521: Cannot read property 'push' of undefined logout error
## Description of Issue

Logging out works but the user still receives an error in the log about push being undefined. It seems the history object is not available to the user in the navbar component.

Possible solution here is to wrap the connect function in the navbar.js file in the withRouter function. This allows the history object to be passed down to the navbar so the user can use history.push to navigate to other pages.

https://medium.com/@jeremylenz/react-router-basics-what-i-wish-theyd-told-me-37b29201d40a

## Additional Comments

```
TypeError: Cannot read property 'push' of undefined
    at Navbar.js:16
    at _callee$ (logout.js:10)
    at tryCatch (runtime.js:45)
    at Generator.invoke [as _invoke] (runtime.js:264)
    at Generator.prototype.<computed> [as next] (runtime.js:98)
    at asyncGeneratorStep (asyncToGenerator.js:3)
    at _next (asyncToGenerator.js:25)

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #521 

## Description

### DIFF
--- a/client/src/components/Navbar/Navbar.js
+++ b/client/src/components/Navbar/Navbar.js
@@ -7,6 +7,7 @@ import kgnuLogo from '../../images/kgnu-logo-white-gray.png';
 import Logo from '../Logo';
 import { authActions } from '../../redux/auth';
 import { bindActionCreators } from 'redux';
+import { withRouter } from 'react-router-dom';
 
 class Navbar extends Component {
   handleSignOut = () => {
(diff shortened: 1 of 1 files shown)


seankwilliams:
@kmid5280 this looks good, nice work! Merging to development.

PR #574, opened 2019-08-30
573 edit

### LINKED ISSUE #573: Dashboard: Set up Edit Event Instance button
## Description of Issue

When there are shows hosted by the current user, you'll see a list of upcoming/past shows on the Dashboard along with text that says `TODO edit show instance link`. Replace this text with a link that says "Edit Show Instance". Clicking the link should open the same modal & form that the "Edit Show Instance" link opens on the Show Calendar.

You may have to bring components over from the Show Calendar, or componentize some things on the Show Calendar that have not yet been componentized.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #573 

## Description

Initial pull request for issue. Would like review and feedback.

### DIFF
--- a/client/src/components/ShowListForUser/ShowListForUser.js
+++ b/client/src/components/ShowListForUser/ShowListForUser.js
@@ -9,6 +9,7 @@ import Loading from '../Loading';
 import { MODAL_EDIT_SHOW } from '../Shows/ShowModalController';
 import { createInstanceAndEditShow } from '../../redux/show';
 import { setModalVisibility } from '../../redux/modal';
+import ShowModalController from '../Shows/ShowModalController';
 
 class ShowListForUser extends Component {
   state = {
@@ -109,6 +110,7 @@ class ShowListForUser extends Component {
                   Edit Show Instance
                 </div>
               </td>
+              <ShowModalController />
             </tr>
           );
         })}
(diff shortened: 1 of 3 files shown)


client/src/components/ShowListForUser/ShowListForUser.js
```diff
@@ -46,7 +61,23 @@ class ShowListForUser extends Component {
...
-  }
+  };
+
+  showEditInstanceModal = show => {
+    const {
+      setModalVisibility,
+      createInstanceAndEditShow,
+      selectShow,
+    } = this.props;
+    if (getShowType(show) === 'instance') {
```
seankwilliams: I added this check to see if an instance was already created. The `createInstanceAndEditShow` function creates an instance from a show series if the instance does not currently exist.

client/src/redux/modal/modalActions.js
```diff
@@ -4,11 +4,12 @@ export const setModalVisibility = (
   modalType,
   modalVisibility,
   data,
+  onModalCloseCallback,
```
seankwilliams: I set up the `setModalVisibility` function to take an optional callback that will fire when the modal is closed. We'll use this so the show instance data refreshes on the dashboard after editing the instance. So, for example, if you change the show's name, now you'll see that change reflected immediately on the dashboard rather than having to reload the page.

 (@jberns  - fyi)

PR #579, opened 2019-09-02
add conditionals to check if track exists in playlist before removal

### LINKED ISSUE #284: Do not allow deleting tracks that are used in Show Builder playlists
Depends on #283
Depends on Show Builder schema being completed

In the same way that #283 made it so artists & albums are not allowed to be deleted, modify the API endpoint for deleting tracks so that tracks that are used in a playlist cannot be deleted.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #284 

## Description

### DIFF
--- a/server/v1/controllers/library/remove.js
+++ b/server/v1/controllers/library/remove.js
@@ -11,7 +11,7 @@ function remove(req, res) {
                 throw 'This album cannot be deleted because it has tracks.';
               }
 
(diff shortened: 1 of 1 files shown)


seankwilliams:
@kmid5280 This wasn't working for me locally so I made a few changes to the promise chain and Mongo queries. Let me know if you have any questions about the changes, merging this now!

PR #584, opened 2019-09-05
354 show builder add track popup

### LINKED ISSUE #354: Show Builder - Add Track popup
depends on #351

When the user clicks the Add button to add a new track, they should see a modal window with the following values:

* Album name
  * Must be selected first. Should be an autocomplete box, similar to the host component in #138.
  * The user should also have the option to add a new album. if they do, they should be prompted for:
    * Artist name (a dropdown that also lets them add a new artist, which would just be a textbox with the artist name)
    * Album name (required)
    * Label (not required)

After providing album, the additional fields are available:
* Track name (required)
* Duration
* Disk number (defaults to the 1, or the maximum disk number for the selected album)
* Track number (defaults to n + 1, where n is the maximum track number for the maximum disk number for the selected album)

There should be a Save button, and clicking it should add the form information into the Music Library database as well as adding the new track to the Scratchpad or Saved Items list.

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #354 

## Description

- Modifies alerts so that they can display within a modal window or outside of a modal window
- Adds support for Library redux to populate an alternate list of documents since the UI in Show Builder requires two lists of libraries (docsForDropdown)
- Wire up an Add Track popup on the Show Builder page


client/src/redux/library/libraryReducers.js
```diff
@@ -57,7 +58,8 @@ export const libraryReducer = (state = initialState, { type, payload }) => {
     case libraryTypes.SEARCH:
       return {
         ...state,
-        docs: payload.docs,
+        docs: payload.useDocsForDropdown ? state.docs : payload.docs,
+        docsForDropdown: payload.useDocsForDropdown ? payload.docs : state.docs,
```
seankwilliams: @dzimmerman902 I came across a situation where I needed two lists of library documents at the same time, so I added a second list here that could be used. Let me know if you have any thoughts on the execution of this! The other relevant piece of this is in the `client/src/redux/library/actions/search.js` file.

client/src/redux/alert/alertReducers.js
```diff
@@ -3,6 +3,7 @@ import { alertTypes } from './alertTypes';
 const initialState = {
   active: false,
   body: '',
+  displayAt: 'main', // either "main" or "modal", this determines the render location for the alert. This will reset whenever alertTypes.INACTIVE is dispatched
```
seankwilliams: @dzimmerman902 I came across a situation where I had to display alerts within a modal window, so I modified Alert to support that. Let me know if you have any feedback or questions on the implementation! I can walk you through this at the meeting on Monday too if that's easier.

PR #594, opened 2019-09-28
587 traffic population issues

### LINKED ISSUE #587: Show Builder: Most traffic events are not populating correctly from the database
## Description of Issue

Go to http://localhost:3000/show-builder/show?startTime=2019-03-13T15:30:00Z&endTime=2019-03-13T18:06:00Z. There are a bunch of traffic events that should be populating, but are not:

missing-traffic-events

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #587 

## Description

Fixes issue where `rrule.between` was not considering Daylight Savings Time when outputting values. Also, fixed problem where times would account for Daylight Savings Time in the server's time zone, not the station's.


server/v1/controllers/shows/utils/utils.js
```diff
@@ -159,16 +162,11 @@ function reduceShowsByRepeatProperty(shows, recurringCheckValue) {
 function returnDatesArrayByRepeatRule(show, startDate, endDate) {
   const rule = new RRule(createRRule(show));
   try {
-    let shows = rule.between(new Date(startDate), new Date(endDate), true);
-
-    //the .between() function will not get a currently playing event, so we'll have to check for it
```
seankwilliams: Removes the use of `before`, as this has the same performance issues that `between` has in the `rrule` library. (The fork I made of rrule only fixes the performance issues for `between`)

PR #599, opened 2019-10-04
163 permissions

### LINKED ISSUE #163: Permissions Schema/API
The Create/Edit user API endpoints should be able to adjust the permissions for each user. The permissions will be exposed in high-level roles like "DJ" or "Music Director", and users can have multiple roles.

**Everyone**
- Can view show calendar, Show Builder, and music library.

**DJ**
•	Can modify their own shows in Show Builder
•	Can modify the description for their own shows
•	No other edit permissions

**Underwriting**
•	Can add/edit traffic calendar
•	Can run reports

**Show Captain** 
•	Can add/edit show calendar 
•	Can add/edit all shows in Show Builder

**Music Library Admin**
•	Can modify records in the music library

**Full Access**
•	Can add/edit everything except users and change Comrad settings (like whether it’s a reporting period)
•	Can run reports

**Admin**
•	Same as Full Access, but can also:
    o	Add/edit users
    o	Change Comrad settings (like whether it’s a reporting period)

Side note: the same permissions will be used for API tokens

### LINKED ISSUE #442: Host Dropdown: Only display the Add New User option if the user has an Admin permission level
Since the Create User API endpoint cannot be used unless a user has a permission level of Admin, the Host Dropdown's Add New User option should not be displayed unless the user has that permission level.

seankwilliams:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #163 
This issue resolves #442 

## Description

Sets up permissions throughout the app, and hides UI elements on the front end when the user doesn't have permissions to use them


cli/seedAccessControl/grants/Hosts.js
```diff
@@ -15,9 +15,9 @@ const Users = [
     attributes: '*',
   },
 
-  // Show Producer
+  // Show Captain
   {
-    role: 'Show Producer',
+    role: 'Show Captain',
```
seankwilliams: @dzimmerman902 I updated the role name here: do you know if there's any test data that needs to be changed for this?

dzimmerman902: @seankwilliams Not that I can think of, should be good!

PR #615, opened 2019-11-17
461 calendar overflow issue

### LINKED ISSUE #461: Show Calendar: Thumbnail calendar has an issue where text temporarily overflows to its right
## Description of Issue

On the Show Calendar page, when moving to previous months on the thumbnail calendar, there is a moment where some of the dates from the previous months appear on the right side of the calendar:

extra-column

Those shouldn't be there, the text from the previous month should completely animate out of sight so there isn't an extra column of numbers.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #461 

## Description

Fixed issue by adding new style override for the Daypicker container in the CalendarHomePage.scss file. Adjusted width of container to prevent overflow.

### DIFF
--- a/client/src/pages/CalendarHomePage/_CalendarHomePage.scss
+++ b/client/src/pages/CalendarHomePage/_CalendarHomePage.scss
@@ -10,6 +10,9 @@
     .DayPicker__horizontal,.CalendarMonth, .CalendarMonthGrid {
       background: none;
     }
+    .DayPicker_transitionContainer {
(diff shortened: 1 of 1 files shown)


kmid5280:
@seankwilliams I think I was able to fix this. Let me know how it looks on your end.

seankwilliams:
@kmid5280 very nice work, this looks great. Merging now!

PR #616, opened 2019-11-18
538 delete

### LINKED ISSUE #538: Make Delete Modal for tracks into a shared component
## Description of Issue

Depends on #476 

Right now, there's some code for the modal asking you to delete tracks that's duplicated on the main Library page and the View Album page. Let's move this code into a reusable component.

While doing this, please remove the delete success modal and instead show a message confirming the delete was successful with `alertActions.show`

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #538 
## Description

Moved Modal code into separate component. Added event binders to constructors in LibrarySearchPage so that DeleteModal component can call functions in parent. Removed DeleteSuccessModal where used and replaced with alert. Currently passing several props into DeleteModal component - may want to look into ways to pass the props in differently, perhaps through the store. Also fixed issue in TableAlbumTracks component where data type was not displaying in delete confirmation dialog.


kmid5280:
@seankwilliams Let me know how this looks. I did add a DeleteSuccessModal component but we can remove that if you aren't going to use it.

client/src/components/DeleteModal/DeleteModal.js
```diff
@@ -0,0 +1,56 @@
...
+          </div>
+        </div>
+      </Modal>
+    );
+  }
+}
+
+function mapDispatchToProps(dispatch) {
+  return {
+    alertActions: bindActionCreators({ ...alertActions }, dispatch),
```
seankwilliams: Since it isn't used, please remove `alertActions` from this file (both in this function and the imports at the top of the file).

client/src/components/DeleteModal/DeleteSuccessModal.js
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react';
```
seankwilliams: Let's remove the `DeleteSuccessModal` - it's function is being replaced by `alertActions.show` (which is more consistent with the rest of the app). The only reason there was a `DeleteSuccessModal` in the first place is because it was written before the alert system was finished.

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
```diff
@@ -34,33 +40,25 @@ class LibrarySearchPage extends Component {
...
-  };
-
-  deleteEntity = (type, id) => {
-    const { libraryActions } = this.props;
-    libraryActions.remove(id, this.deleteSuccess, this.deleteFailure);
-  };
-
   deleteFailure = () => {
     window.scrollTo(0, 0);
+    console.log('error');
```
seankwilliams: Can you either remove this `console.log` message or make it a little more descriptive so it explains the error a little bit? Since this error will be appearing from a compiled React app once this is in production, the error message should be unique enough that you could find it by searching the files for that particular error message (in case the app is compiled in a way where the end user cannot see the individual files and line numbers where error messages were generated). Or, ideally, you should be able to tell what the message means if you just saw it in a list of console messages.

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
```diff
@@ -34,33 +40,25 @@ class LibrarySearchPage extends Component {
...
-      function() {
-        //refresh data from the database - https://github.com/tannerlinsley/react-table/issues/808#issuecomment-373673915
-        this.table.fireFetchData();
-      },
+    this.props.alertActions.show(
+      'success',
+      'Success',
+      `${entity.data.name} was successfully deleted`,
     );
+    this.setState(function() {
```
seankwilliams: Is `this.setState` necessary since the state is no longer being changed?

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
kmid5280: I'll go ahead and remove it for now, since I added it in just to see if it was being triggered and just forgot to take it back out. However is there a particular way you would suggest throwing that error? Would you recommend just having a more detailed console message such as 'Error deleting file,' or is there a better way?

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
kmid5280: I'm not sure what that does. I didn't move it over to the component though. What do you recommend?

seankwilliams: @kmid5280 `setState` is a function used by React to update the state: https://css-tricks.com/understanding-react-setstate/. .Since you aren't passing any values for state changes, I would remove `this.setState` and only leave the call to `this.table.fireFetchData()`.

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
seankwilliams: @kmid5280 Thanks! I think removing it is just fine and I don't think anything is necessary here: There is an alert being shown telling the user the reason the delete failed (I believe that's done in the Redux action), and I don't think there's any extra information a developer would need to know in the JavaScript console.

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
kmid5280: Oh I know what setState does, I just wasn't sure what role it was playing in this particular case. If it's not needed I can go ahead and remove it.

kmid5280:
@seankwilliams That should take care of everything. Let me know if there are any issues.

client/src/pages/LibrarySearchPage/LibrarySearchPage.js
seankwilliams: @kmid5280 Ah okay sorry I misunderstood there. Looking at the code that's being removed in the PR, setState was previously setting this in state:
```
     {
        deleteSuccessModal: entity.data,
      },
```
And since we've removed the success modal, setting the data for it is no longer necessary.

seankwilliams:
@kmid5280 Thank you, this is looking great! Merging now.

I made two commits that are code cleanup things. The first was to fix these warnings about undefined variables:

old-code

And, the second was to remove these `bind` functions, which aren't necessary because the `closeDeleteModal`, `deleteSuccess` and `deleteFailure` functions are defined with arrow functions: which automatically takes care of setting `this` in the context of those functions. If we had defined those functions like `function closeDeleteModal() { ... }`, then the bind code would be necessary:

old-code-2

PR #630, opened 2020-03-12
629 section

### LINKED ISSUE #629: Users section changes
**On the "View User" page (click the Users icon in the left menu and click on any user's name):**
Rearrange the section called "Admin" in the left column:
- [x] Change the word "Admin" to "Access and API"
- [x] Display "Active" and "Inactive" for "Status" rather than "TRUE"/"FALSE", and change this to pull from the "status" field rather than representing whether the user has an API key
- [x] Remove the "Can Delete" property (this is only for use in validating deleting a user)
- [x] If `can_delete` is true, display the DELETE USER button
- [x] If `can_delete` is false, display the text "This user cannot be deleted because they are a host of a show."
- [x] Add a subheading at the bottom called "API Key". If the user has an API key, display "User has API key", the "Reset" button, and the "Delete" button. Change the wording to "Reset API Key" and "Delete API Key". If the user does not have an API Key, display a button saying "Create API Key"

In the middle column:
- [x] In Firefox, the default user profile icon displays very small. Please add CSS so this displays the same size it does in Chrome:
image

In the right column:
- [x] Display a list of the user's past shows and upcoming shows. We can use the same component used on the Dashboard for this.

**On the Users List/Search Page (you get to this by clicking the Users icon on the left):**
- [x] Add a three dots icon that opens a menu with "Edit" to each row: similar to what is on the Library Search Page.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #629 

## Description


kmid5280:
I changed the style in the Heading component in an attempt to fix a centering issue. The center property was defined using text-align: center, but this was not centering the header. By changing it to `left: 0; right: 0: margin: auto`, it will center, but may want to make sure this doesn't affect the Header component negatively anywhere else in the app.

seankwilliams:
@kmid5280 this is all looking great, thank you. I made a commit that cleaned up a few small issues: the main thing is I changed the API & Access card so it's not using the `<Row>` component because that was wrapping elements in a div with `margin-left:-15px`, which was making the padding with that card inconsistent with the other cards on the page. I replaced the `Row` elements with `div` and it seems to have kept the styles (as a bonus, this means the .Heading--center class can stay as `text-align: center`). Let me know if you have any questions about the changes. Merging this in now!

PR #663, opened 2020-08-08
getting_started.md: Add note on specifying dotenv file if it's not

jessicaschilling:
## Description

Minor tweak to `getting_started.md` to account for error I encountered on clean install.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -37,4 +37,10 @@ If, on using `npm run [command]`, you receive an error like the following:
 [1] However, a different version of babel-loader was detected higher up in the tree:
 ```
 
-Then, open/create the `.env` file in the client folder and add the line `SKIP_PREFLIGHT_CHECK=true`
+Then, open/create the `.env` file in the client folder and add the line `SKIP_PREFLIGHT_CHECK=true`.
+
+If you have to create a `.env` file in your client folder, you may also need to specify its location in the `server/index.js` file to avoid a Mongoose error. Edit line 1 as so:
+
+```
+require('dotenv').config({ path: 'client/.env' });
+```


seankwilliams:
Thank you @jessicaschilling ! Were you able to successfully launch the project after this change? There might be another missing step that's the culprit here. There should be two `.env` files: one at the root of the project, and one at `client/.env`. I haven't updated the "Getting Started" instructions for this yet though. I will update those instructions tomorrow and let you know.

Also, if you're interested in contributing, please let me know if there's a particular part of the stack you'd like to work in (back-end, front-end, etc.) I can go through a few issues and offer suggestions, as well as being sure the issues have enough detail to start on.

jessicaschilling:
Hey @seankwilliams - thanks for the update! I was able to get the front end running but ended up with a dashboard full of loading spinners - that said I'm really only a front-end person so that could be user error.

I'm afraid I've got about zero discretionary time these days but I'll have a nosey through the issues just in case! Thanks.

seankwilliams:
@jessicaschilling I've revised the Getting Started instructions so they should work better! The new version is at https://github.com/codefordenver/Comrad/blob/development/docs/getting_started.md. Steps 2-4 are new, so you'll have to run through those. You'll want to replace any existing .env files you have. Could you give those a try and see if that gets the project running for you?

jessicaschilling:
@seankwilliams Appreciate the extra instructions - these are clear but I'm getting hung up on setting up MongoDB. Not your problem -- thanks though!

seankwilliams:
@jessicaschilling You bet. Please get the latest from the `development` branch when you do get Mongo working. I just made some schema changes to the project.

PR #693, opened 2020-12-03
add onSelecting prop to component

### LINKED ISSUE #676: Sometimes, clicking on calendar opens instance instead of the tooltip (usually on initial load)
On the Show Calendar, sometimes, clicking on an existing calendar event opens a window for creating a new show rather than opening the tooltip to modify the existing calendar event.

I believe this is actually due to users slightly dragging the mouse button up and down as they click, which the show calendar interprets as the user wanting to create a new event covering the timespan the mouse has been dragged across. If there is a way to make the show calendar a little less sensitive to the mouse being dragged up and down, that may resolve this issue.

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #676 

## Description

Try this solution and see if it works. Added onSelecting prop to component set to false. However this may prevent dragging and dropping of existing shows.

### DIFF
--- a/client/src/components/Shows/ShowCalendar.js
+++ b/client/src/components/Shows/ShowCalendar.js
@@ -238,6 +238,7 @@ class Calendar extends Component {
           components={{
             eventWrapper: this.customEventWrapper,
           }}
+          onSelecting={slot => false}
         />
 
         <ShowModalController />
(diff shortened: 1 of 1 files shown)


seankwilliams:
@kmid5280 this is working great! And that's just fine if dragging & dropping shows is disabled, because if you could do that before, it wouldn't have worked properly anyways. I'll merge this in, thank you!

PR #713, opened 2021-04-26
Fix for traffic.traffic_details rendering errors

### LINKED ISSUE #712: Adding new giveaway in traffic page without specifying event name causes error
## Description of Issue

If you add a new Giveaway in the traffic page and don't specify an event name, then the page will return an error when you attempt to view the time slow that the Giveaway is in in Show Builder.

To replicate:
1. Go to the Traffic page.
2. Click Add.
3. Select 'Giveaway' from the Type dropdown menu.
4. Create a new Giveaway without specifying an event name.
5. Go to Show Builder.
6. Click on the time slot where the Giveaway will appear in the scratchpad. This will return an error, and you will not be able to see any of the information for that time slot.

Make the Event Name a required field, or replace with 'undefined' if no entry is given.

## Additional Comments

_Comments Go Here_

kmid5280:
## Guidelines

YOUR branch on the RIGHT

## Issue

This issue resolves #712 

## Description

Added conditionals to check for existence of `traffic.traffic_details.description` and `traffic.traffic_details.giveaway_details` properties before rendering.

### DIFF
--- a/client/src/components/ShowBuilderItemList/ShowBuilderItemList.js
+++ b/client/src/components/ShowBuilderItemList/ShowBuilderItemList.js
@@ -43,6 +43,7 @@ class ShowBuilderItemList extends Component {
 
     if (
       'giveaway' in configState.customFields &&
+      traffic.traffic_details.giveaway_details != null &&
       traffic.traffic_details.giveaway_details.custom != null
     ) {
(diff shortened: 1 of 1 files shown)


seankwilliams:
Beautiful. Thank you @kmid5280 !

PR #837, opened 2023-09-03
Fix typo in Wednesday

kmid5280:
## Description

Fixed a typo in a few files. This doesn't appear to be a breaking change but feel free to double check.


seankwilliams:
Thanks for catching that, @kmid5280 ! This shouldn't cause any issues.

PR #866, opened 2023-10-27
863 defaultdate

### LINKED ISSUE #863: Change default dates for Sound Exchange and Underwriting reports
Let's change the pages at:

/reporting/sound-exchange
/reporting/underwriting

So that the default date range has a from date of 30 days ago, and a to date of today.

kmid5280:
## Issue

This issue resolves #863 

## Description

Give this a try. Note: The initial value for the "From" field might be off by a day, which can be fixed by changing the `priorDay` variable in FormDateRangeForExport to -31 instead of -30.

### DIFF
--- a/client/src/components/forms/FormDateRangeForExport/FormDateRangeForExport.js
+++ b/client/src/components/forms/FormDateRangeForExport/FormDateRangeForExport.js
@@ -9,12 +9,34 @@ import { DatePicker__React } from '../../DatePicker';
 import DropdownUnderwriter from '../../DropdownUnderwriter';
 
 class FormDateRangeForExport extends Component {
+  constructor(props) {
+    super()
+    this.state = {
+      // These values are for setting initial values for input dates if page requires it.
+      today: '',
+      priorDay: '',
+      useInitialDateValues: props.useInitialDateValues, // This is passed in from page to determine whether initial date values need to be set in input fields.
+    }
+  }
+  
   submit = values => {
     const { submitCallback } = this.props;
     if (typeof submitCallback === 'function') {
       submitCallback(values);
     }
   };
+
+  componentDidMount() {
+    if (this.state.useInitialDateValues) {
+      const today = new Date()
+      const priorDay = new Date()
+      priorDay.setDate(priorDay.getDate() - 30)
+      this.setState({
+        today: today,
+        priorDay: priorDay,
+      })
+    }
+  }
 
   componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


seankwilliams:
@kmid5280 Those defaults are looking good, but if you select a different date on the datepicker, the field doesn't update. Can you take a look?

kmid5280:
@seankwilliams I made some updates here. I moved the initial date values to the component state, set them via the ComponentDidMount lifecycle hook, and changed the method for updating them using the onChange attribute in the input fields. Let me know how that looks to you.

seankwilliams:
@kmid5280 That's looking good, nice work. I'll get this merged in the next time I do a deployment to the live Comrad site.

seankwilliams:
@kmid5280 This is live now, and while the default dates are appearing in the fields, they aren't being passed along when submitting the form and generating the report. You have to change the date values for them to be used when submitting the form. Can you take a look?

kmid5280:
@seankwilliams Thanks for pointing this out. I'm sorry this isn't working right. Hm. I spent some time on this and am struggling to find a way to get the DatePicker component to automatically set an initial date value. Currently, no date value is being registered in the component unless the date is manually selected in the DatePicker popper, even though the date is being displayed correctly in the input field. I can confirm this by using console.log in FormDateRangeForExport in the submit method to look at the values being passed in.

I tried setting the initial value in the state just to test it, bypassing the need for the componentDidMount lifecycle method.

I'm seeing this issue referenced in some discussion forums, though it's not clear to me what the solution is. https://github.com/Hacker0x01/react-datepicker/discussions/3861

Are there any instances in Comrad where the DatePicker sets an initial value on loading and works correctly? I'm wondering if we may be running into a technical limitation with this plugin.

seankwilliams:
@kmid5280 It's possible that you'll need to look at the `redux-form` component that sets up the form for this page since that is controlling the values used for the date picker.

The Add Traffic page has values for the default of the date picker: try navigating to `/traffic/add/2023-11-06T00:40:00-06:00` for example and you'll see a value prepopulated for the date picker field.

kmid5280:
@seankwilliams I'm looking at the FormTraffic page to see how it's structured there. What if in FormDateRangeForExport, we define the initial date values via mapStateToProps and pass those into the input fields from there?

seankwilliams:
@kmid5280 Let's try it and see, that seems like it's the right move! I honestly always forget how Redux Form works, so I can't remember for sure.

PR #870, opened 2023-11-08
Fix initial date values on FormDateRangeForExport

kmid5280:
## Issue

This issue resolves #866 

## Description

This form was not correctly initializing the From and To dates when opened. Moved the conditional logic to set the initial date values into mapStateToProps.


kmid5280:
@seankwilliams Give this a try. I tested this fix and it seems to work, though I am a little unfamiliar with using `initialValues` in Redux Form. Might want to double check the usage of props vs state in this form just to make sure everything lines up okay.

seankwilliams:
This looks great. Nice work, @kmid5280 !

PR #886, opened 2023-11-15
Added new values to ticket type field

### LINKED ISSUE #876: Ticket giveaway form changes
Add the following values for the ticket type field of the giveaway form:

Digital - AXS
Digital - Ticketmaster
Will Call
PDF tickets

Also, change the ticket quantity field to default to a value of 2.

kmid5280:
## Issue

This issue resolves #876 

## Description

Added new values to the ticket type field. It looks as though the default value for the Ticket Quantity field is already set to 2.


seankwilliams:
Thank you @kmid5280 ! I'll coordinate with KGNU to get this out. They'll have to make some changes on their end at the same time this goes live - the values from ticket giveaways get passed over to Jotform, so they have to add the same ticket types as options there.

PR #888, opened 2023-11-16
Increase padding value for Rich Text Editor

### LINKED ISSUE #887: Padding issue in Add Traffic description box
## Description of Issue

It looks like the description box in the Add Traffic form could use additional padding.

We could change the padding value for the `.ql-editor` attribute in the RichTextArea scss file, though this would affect all pages where the RichTextArea component is used.

traffic-add-desc

kmid5280:
## Issue

This issue resolves #887 

## Description

Increased padding value for inside of Rich Text Editor.


seankwilliams:
Thanks @kmid5280 , I'll get this out when I do the next deployment!

PR #896, opened 2024-03-07
Add old library number to search index

### LINKED ISSUE #875: Add ability to search library by old 5 digit library number
Change the library search so, in addition to the current fields it looks at, it will also consider the value at custom.library_number

kmid5280:
## Issue

This issue resolves #875 

## Description

Added conditional to include album library number to track search index.


seankwilliams:
@kmid5280 I tested this and it's working great. Deploying this now. Thanks for the hard work on this one!

